PR #1, opened 2023-08-31
podspec

rkistner:
Wrap the xcframework in a powersync-sqlite-core pod for easy installation.

Also fixes some issues around _Unwind_Resume causing crashes on iOS.

TODO:
 * [x] Fix issues with the podspec to allow publishing publically.


stevensJourney (approved):
LGTM

PR #19, opened 2024-08-12
Sync Optimizations

rkistner:
1. Remove handling of `{target: ...}` in MOVE operations. This has been removed from the protocol, and has no active
implementations. See [protocol docs for MOVE](https://github.com/powersync-ja/powersync-service/blob/0c253a634abdfcce4185d9ce126f73ea0b682293/docs/sync-protocol.md?plain=1#L154).
2. Never persist MOVE operations - we only care about the checksum.
3. Do not persist REMOVE operations if any of these hold:
   1. This is an initial sync for the bucket (including adding new buckets). However, these do still need to supersede previous PUT operations.
   2. There was no previous operation supserseded (REMOVE for a PUT operation we don't have).
   3. More generally, if the only superseded operations were not applied locally yet, meaning there is nothing to remove locally.
4. Fix a crash due to wrapping checksums (integer overflow) in debug builds. Release builds did not have this issue.
5. When receiving a new operation for a row, instead of marking the previous operation as superseded, delete it.

Combined, these optimizations could help to significantly speed up initial sync of compacted buckets, where a large percentage of operations are MOVE or REMOVE operations.

This will not have a significant performance impact if:
1. The bucket is not compacted at all (meaning no MOVE operations).
2. The bucket is fully defragmented (meaning only PUT operations).

## Benchmark 1 - many MOVE and REMOVE ops

Test case:
 * Local powersync-service
 * 122,422 total operations (30MB downloaded), with:
   * 2,422 PUT operations (13MB of data)
   * 60,000 MOVE operations
   * 60,000 REMOVE operations

### Dart native, Linux desktop

Before: 5.7s
After: 2.7s

### Diagnostics app (web sdk)

Disabled dynamic schema generation.

Before: 74s for saving the data, 490s (!) for compacting
After: 7.8s

The big speedup is likely from not filling ps_oplog with many REMOVE operations. Will need further investigation to determine why it was so slow, since we may still get this kind of performance for other cases with a large number of operations.

## Benchmark 2 - many PUT ops for small number of rows

Test case:
 * Local powersync-service
 * 60k total operations (21MB downloaded), over 20 rows

### Dart native, Linux desktop

Before: 2.34s, 800KB db file, 4.6MB WAL
After: 2.07s, 4KB db file, 2.6MB WAL

This gives around 10% performance improvement, but with significantly reduced storage usage.

### Diagnostics app (web sdk)

Before: 77s, 2.2MB storage
After: 37s, 4.5MB storage

It's not clear why the storage increased in this case.

### Remove superseded column

We can completely remove the `superseded` column - it is now always 0. This will require some semi-tricky migrations, so we're not doing it just yet (there may be existing data where `superseded = 1`).

### Optimize compacting

Now that superseded operations are immediately deleted, we can also optimize the compact operations (clear_remove_ops). By combining this with the `SET last_applied_op = last_op` part, we can significantly reduce the number of rows we need to scan for REMOVE operations after incremental updates. This can give us continuous auto-compacting, instead of the current "compact once every 1000 operations".

### Normalize bucket names

Bucket names are primarily used when saving and superseding operations. We already store each synced bucket in `ps_buckets`. We can use those ids in `ps_oplog`, instead of the full bucket names. This could reduce storage size and increase performance.

### DIFF
--- a/crates/core/src/checkpoint.rs
+++ b/crates/core/src/checkpoint.rs
@@ -1,8 +1,7 @@
 extern crate alloc;
 
-
 use alloc::format;
-use alloc::string::{String};
+use alloc::string::String;
 use alloc::vec::Vec;
 use core::ffi::c_int;
 use core::slice;
@@ -34,7 +33,8 @@ fn powersync_validate_checkpoint_impl(
     let db = ctx.db_handle();
 
     // language=SQLite
-    let statement = db.prepare_v2("WITH
+    let statement = db.prepare_v2(
+        "WITH
 bucket_list(bucket, lower_op_id, checksum) AS (
   SELECT
         json_extract(json_each.value, '$.bucket') as bucket,
@@ -57,7 +57,8 @@ FROM bucket_list
      bucket_list.bucket = oplog.bucket AND
      oplog.op_id <= CAST(json_extract(?1, '$.last_op_id') as INTEGER) AND
      oplog.op_id > bucket_list.lower_op_id
-GROUP BY bucket_list.bucket")?;
+GROUP BY bucket_list.bucket",
+    )?;
 
     statement.bind_text(1, data, sqlite::Destructor::STATIC)?;
 
@@ -65,6 +66,7 @@ GROUP BY bucket_list.bucket")?;
 
     while statement.step()? == ResultCode::ROW {
         let name = statement.column_text(0)?;
+        // checksums with column_int are wrapped to i32 by SQLite
         let add_checksum = statement.column_int(1)?;
         let oplog_checksum = statement.column_int(2)?;
         let _count = statement.column_int(3)?;
@@ -72,7 +74,8 @@ GROUP BY bucket_list.bucket")?;
         let _last_applied_op = statement.column_int64(5)?;
         let expected_checksum = statement.column_int(6)?;
 
-        let checksum = oplog_checksum + add_checksum;
+        // wrapping add is like +, but safely overflows
+        let checksum = oplog_checksum.wrapping_add(add_checksum);
 
         if checksum != expected_checksum {
             failures.push(String::from(name));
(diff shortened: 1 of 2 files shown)


crates/core/src/checkpoint.rs
```diff
@@ -57,22 +57,25 @@ FROM bucket_list
...
         let add_checksum = statement.column_int(1)?;
         let oplog_checksum = statement.column_int(2)?;
         let _count = statement.column_int(3)?;
         let _last_op_id = statement.column_int64(4)?;
         let _last_applied_op = statement.column_int64(5)?;
         let expected_checksum = statement.column_int(6)?;
 
-        let checksum = oplog_checksum + add_checksum;
+        // wrapping add is like +, but safely overflows
+        let checksum = oplog_checksum.wrapping_add(add_checksum);
```
rkistner, line 78: Rust default overflow semantics for i32 is panic in debug mode, and do a wrapping add in release mode. This is why it previously worked in release mode, but could crash hard in debug mode.

`wrapping_add` fixes this to work correctly in both debug and release mode.

crates/core/src/operations.rs
```diff
@@ -151,77 +161,60 @@ UPDATE ps_buckets
...
-        statement.bind_text(2, bucket, sqlite::Destructor::STATIC)?;
-        statement.bind_int64(1, *last_op)?;
-        statement.exec()?;
-    }
-
-
-    // Compact superseded ops immediately
-    if let (Some(first_op), Some(last_op)) = (&first_op, &last_op) {
-        // language=SQLite
-        let statement = db.prepare_v2("UPDATE ps_buckets
```
rkistner, line 173: This bit is now largely redundant, since we don't write superseded MOVE/REMOVE ops directly anymore. Technically it is still possible to get a PUT op and supersede it in the same batch, but we just compact those later as part of `clear_remove_ops`.

stevensJourney (approved):
These are some amazing performance improvements.

rkistner:
Will get a new release with these changes out next week.

Note that for the web SDK, the changes in https://github.com/powersync-ja/powersync-js/pull/266 fix the biggest performance issues, but these changes will give further improvements for buckets with many more operations than actual rows.

PR #21, opened 2024-08-20
powersync_client_id() / powersync_last_synced_at() / powersync_clear()

rkistner:
Add support for:

### `powersync_client_id()`

This returns an unique uuid per client. This is configured once when initializing the powersync tables.

This allows us to:
1. Identify individual syncing clients, instead of just users. This will make it easier to notice issues such as multiple concurrent connections for the same client.
2. Track write checkpoints per client instead of per user.

A new optional parameter will be added to the protocol for this.

### `powersync_last_synced_at()`

This returns a persisted timestamp of the last synced at time. This should replace the `SELECT 1 FROM ps_buckets WHERE last_applied_op > 0 LIMIT 1` query currently used in clients, and will work even when no data was synced.

### `powersync_clear()`

This clears the database. `select powersync_clear(1)` to delete everything, `select powersync_clear(0)` to preserve local tables.

For the new ps_kv table, this clears last_synced_at, but not client_id.

This contains a workaround to trigger one update notification for each table, while still using the sqlite truncate optimization for the majority of rows.

This should replace the delete logic that is currently implemented in each client separately.

## Internals

This now implements down migrations, which allows a developer to downgrade the SDK to older versions.

`powersync_client_id` and `powersync_last_synced_at` are both stored in a new `ps_kv` table.

Tests for the new functionality are implemented here: https://github.com/powersync-ja/powersync.dart/pull/148

This PR removes lots of unused (incomplete) code, and fixes some lint issues.


mugikhan (approved):
Looks good!

PR #24, opened 2024-08-22
Custom prefab build for Android

rkistner:
When we just want to bundle the `libpowersync.so` files into the final .apk, the `jniLibs` approach works file. However, if we want to use the libs from another native Android library, such as with `react-native-quick-sqlite`, that approach is insufficient. Instead, we need [Prefab](https://developer.android.com/build/native-dependencies) for that.

The native Android tooling only supports building Prefab libraries using CMake or the NDK. Lucky for us, the Prefab format is very simple, and we can create a custom .aar file with the Prefab structure. We now completely avoid the android gradle plugin for generating the aar file.

The final .aar file will contain the following files:

```
AndroidManifest.xml
jni/arm64-v8a/libpowersync.so
jni/armeabi-v7a/libpowersync.so
jni/x86_64/libpowersync.so
jni/x86/libpowersync.so
META-INF/com/android/build/gradle/aar-metadata.properties
prefab/modules/powersync/include/powersync.h
prefab/modules/powersync/libs/android.arm64-v8a/abi.json
prefab/modules/powersync/libs/android.arm64-v8a/libpowersync.so
prefab/modules/powersync/libs/android.armeabi-v7a/abi.json
prefab/modules/powersync/libs/android.armeabi-v7a/libpowersync.so
prefab/modules/powersync/libs/android.x86_64/abi.json
prefab/modules/powersync/libs/android.x86_64/libpowersync.so
prefab/modules/powersync/libs/android.x86/abi.json
prefab/modules/powersync/libs/android.x86/libpowersync.so
prefab/modules/powersync/module.json
prefab/prefab.json
```

Usage in the consuming project:

```gradle
// build.gradle
dependencies {
    implementation 'co.powersync:powersync-sqlite-core:0.2.0'
}
```

```
# CMakeLists.txt
find_package(powersync_sqlite_core REQUIRED CONFIG)

target_link_libraries(
  ${PACKAGE_NAME}
  # ...
  powersync_sqlite_core::powersync
)
```

The `prefab` folder is new, the rest were there before.

Note: prefab is _only_ relevant when linking to this library via native code on Android. FFI or loading the extension via SQLite APIs does not need this.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -52,3 +52,26 @@ rustflags = [
     "-C", "link-arg=-mmacosx-version-min=10.13",
 ]
 
+
+# For Android, it is important to set the soname.
+# Otherwise, the linker hardcodes the path in the lib,
+# which breaks loading.
+[target.aarch64-linux-android]
+rustflags = [
+    "-C", "link-arg=-Wl,-soname,libpowersync.so",
+]
+
+[target.armv7-linux-androideabi]
+rustflags = [
+    "-C", "link-arg=-Wl,-soname,libpowersync.so",
+]
+
+[target.x86_64-linux-android]
+rustflags = [
+    "-C", "link-arg=-Wl,-soname,libpowersync.so",
+]
+
+[target.i686-linux-android]
+rustflags = [
+    "-C", "link-arg=-Wl,-soname,libpowersync.so",
+]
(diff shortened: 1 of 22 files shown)


stevensJourney (approved):
Awesome to have prefab now. This will make the use in React Native Quick SQLite much nicer.

mugikhan (approved):
This will make integration with other libraries such as op-sqlite in the future much easier!

PR #26, opened 2024-09-09
Checksum Caching

rkistner:
Cache checksums to improve performance when syncing 100k+ rows.

## Benchmarks

Flutter app, Linux desktop.

Start with 500k PUT operations, then time the latency of syncing 1 PUT or REMOVE operation.

Latency is timed from the client receiving the new checkpoint, to the changes being fully applied locally.

Before: ~860ms
After: ~3ms
After, REMOVE operation: ~300ms

REMOVE operations are still expensive due to the compact performed afterwards. The compact can be performed only if there are a high number of REMOVE operations, which will amortize the cost (added as a new option to the `clear_remove_ops` operation).

### DIFF
--- a/crates/core/src/view_admin.rs
+++ b/crates/core/src/view_admin.rs
@@ -131,7 +131,7 @@ CREATE TABLE IF NOT EXISTS ps_migration(id INTEGER PRIMARY KEY, down_migrations 
         return Err(SQLiteError::from(ResultCode::ABORT));
     }
 
-    const CODE_VERSION: i32 = 3;
+    const CODE_VERSION: i32 = 4;
 
     let mut current_version = current_version_stmt.column_int(0)?;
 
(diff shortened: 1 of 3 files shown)


crates/core/src/view_admin.rs
```diff
@@ -250,6 +250,26 @@ INSERT INTO ps_migration(id, down_migrations) VALUES(3, json_array(json_object('
...
+        // language=SQLite
+        local_db.exec_safe("\
+ALTER TABLE ps_buckets ADD COLUMN op_checksum INTEGER NOT NULL DEFAULT 0;
+ALTER TABLE ps_buckets ADD COLUMN remove_operations INTEGER NOT NULL DEFAULT 0;
+
+UPDATE ps_buckets SET op_checksum = (
+  SELECT IFNULL(SUM(ps_oplog.hash), 0) & 0xffffffff FROM ps_oplog WHERE ps_oplog.bucket = ps_buckets.name
+);
+
+INSERT INTO ps_migration(id, down_migrations)
```
stevensJourney, line 263: I see this table is only introduced in https://github.com/powersync-ja/powersync-sqlite-core/pull/27/files. Which is probably okay if this is released at the same time.

stevensJourney: My bad. I was just about to close this after seeing that :)

PR #28, opened 2024-09-11
Create internal views on `powersync_init()`, not extension init

rkistner:
Some temporary internal views were created when the extension is initialized on a database.

This causes issues when using the extension with SQLCipher: The `PRAMGA key = '...'` command must be executed before the database file is touched, but that's not possible when the views are created when the extension is initialized.

This moves the view creation from the extension initialization to the `powersync_init()` function.


mugikhan (approved):
LGTM, looking forward to testing this with SQLCipher.

PR #34, opened 2024-10-08
Build fixes and improvements

rkistner:
This now adds explicit `-Z build-std` and `--target` specifiers to every release build.

The significant effects:
1. The `--target` ensures we actually build for the correct target. The current linux `libpowersync_aarch64.so` release is actually a x86_64 build.
2. build-std gives us a more optimized std-lib, and specifically removes any unwinding-related code. This reduces the output library size (241KB -> 173K for linux x86_64),
3. The above avoids `undefined symbol: _Unwind_Resume` or similar issues.

build-std also requires an explicit target. 

Additionally, this fixes some linux targets incorrectly specified in `.cargo/config.toml`.

This specifically affects linux, macos and windows builds - will have to test each of those again. Android, iOS and WASM already used build-std, and are not affected.

This also fixes some issues in the release process:
 * Only upload binaries in the release.yml workflow - the others were not used.
 * Use `gh release upload` for binary uploads. The previous method created both a draft release and another published release; this now only creates a single draft.

Sizes of artifacts after these changes:
image

Note: Manually rebuilt libpowersync_aarch64.so for release 0.2.0, 0.2.1 and 0.3.0 to fix the build issue in current releases.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,24 +1,35 @@
-# On Linux, if we don't link to gcc_eh, we get can get this error when loading the loadable extension:
-# undefined symbol: _Unwind_Resume
-# This adds around 29KB to the loadable extension.
-# It may also be an option to just define _Unwind_Resume, but it causes crashes on errors on e.g. iOS, so rather avoid it.
+
+# Previously we added this to rustflags for all linux builds:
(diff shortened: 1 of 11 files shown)


mugikhan (approved):
This makes the build process much simpler!

PR #35, opened 2024-10-10
WASM builds

rkistner:
This adds a side-module WASM build. After lots of back and forth, it turns out the simplest way to get this working is to _remove_ workarounds used before. This now builds the wasm file directly using rust - no custom emcc calls required.

For static build for sqlite3, this now uses a `staticlib` crate-type, producing libpowersync.a. This also removes the need of the workarounds used before with the .bc files, and can significantly simplify the sqlite3 build.

See for more context on the dynamic lib: https://github.com/powersync-ja/wa-sqlite/pull/16

Artifacts included in the release:
 * `powersync.wasm` - synchronous side-module, emscripten-only
 * `powersync-async.wasm` asyncify side-module, emscripten-only
 * `libpowersync-wasm.a` (static lib, works for synchronous emscripten, asyncify empscripten and plain wasi builds)

Example artifacts: image


stevensJourney (approved):
LGTM!

rkistner:
Manually added the builds to v0.3.0 release.

PR #55, opened 2025-01-30
Support bucket with different priorities

simolus3:


crates/core/src/sync_local.rs
```diff
@@ -80,21 +98,30 @@ FROM updated_rows b
...
+        let buckets = statement.column_int(4)?;
+        let data = statement.column_text(3);
 
         let table_name = internal_table_name(type_name);
 
+        if local && buckets == 0 && priority == BucketPriority::HIGHEST {
+            // These rows are still local and they haven't been uploaded yet (which we allow for
+            // buckets with priority=0 completing). We should just keep them around.
+            continue;
+        }
```
simolus3, line 123: This one definitely needs to be checked for correctness. The idea is that clients would call `sync_local` for each partially-completed checkpoint they've received from the sync service. Calling `sync_local` with the lowest possible priority is equivalent to a fully completed checkpoint.

So for a partial checkpoint with the highest priority, we just keep all the local data (and lower priorities would not get to this point with entries in the upload queue). It looks like there might be a potential race condition here if an upload (eventually deleting `ps_updated_rows` if I understand correctly) runs concurrently with a sync session deliverying a partial checkpoint with priority 0?

simolus3 (commented):
Adding some of my own comments for a discussion on this.

crates/core/src/operations.rs
```diff
@@ -17,8 +18,11 @@ SELECT
     json_extract(e.value, '$.data') as data,
     json_extract(e.value, '$.has_more') as has_more,
     json_extract(e.value, '$.after') as after,
-    json_extract(e.value, '$.next_after') as next_after
-FROM json_each(json_extract(?, '$.buckets')) e",
+    json_extract(e.value, '$.next_after') as next_after,
+    json_extract(d.value, '$.priority') as priority
+FROM json_each(json_extract(?1, '$.buckets')) e
+    LEFT OUTER JOIN json_each(json_extract(?1, '$.descriptions')) d
+        ON json_extract(e.value, '$.bucket') == d.key",
```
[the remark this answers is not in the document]
rkistner: Could we perhaps include the priority inside the buckets json array, instead of a separate field?

I'm also wondering, do we need to persist these priorities at all? If the app or connection is ever restarted, we'll always get new bucket priorities before we get a "partial checkpoint complete" message, so it could work to just keep the priorities in memory. That would then mean passing an array of buckets in with the `sync_local` operation, instead of / in addition to the priority. I haven't fully thought through this and whether there are other implications though.

crates/core/src/sync_local.rs
```diff
@@ -61,16 +89,22 @@ pub fn sync_local(db: *mut sqlite::sqlite3, _data: &str) -> Result<i64, SQLiteEr
...
+    SELECT DISTINCT FALSE as local, b.row_type, b.row_id FROM ps_buckets AS buckets
+      CROSS JOIN ps_oplog AS b ON b.bucket = buckets.id AND (b.op_id > buckets.last_applied_op)
+      WHERE buckets.id IN (SELECT id FROM involved_buckets)
+    UNION SELECT TRUE, row_type, row_id FROM ps_updated_rows
+  )
 
 -- 3. Group the objects from different buckets together into a single one (ops).
 SELECT b.row_type as type,
     b.row_id as id,
+    b.local as local,
```
rkistner, line 107: Due to the GROUP BY row_type, row_id, this could select an arbitrary value if the row is present in both ps_oplog and ps_updated_rows.

crates/core/src/sync_local.rs
```diff
@@ -61,16 +89,22 @@ pub fn sync_local(db: *mut sqlite::sqlite3, _data: &str) -> Result<i64, SQLiteEr
...
+WITH 
+  involved_buckets (id) AS (
+    SELECT id FROM ps_buckets WHERE ?1 IS NULL
+      OR name IN (SELECT value FROM json_each(json_extract(?1, '$.buckets')))
+  ),
+  updated_rows AS (
+    SELECT DISTINCT FALSE as local, b.row_type, b.row_id FROM ps_buckets AS buckets
+      CROSS JOIN ps_oplog AS b ON b.bucket = buckets.id AND (b.op_id > buckets.last_applied_op)
+      WHERE buckets.id IN (SELECT id FROM involved_buckets)
+    UNION SELECT TRUE, row_type, row_id FROM ps_updated_rows
```
rkistner, line 101: `ps_updated_rows` is currently used to track both:
1. Changes made locally (to allow reverting the changes if the server does not apply them).
2. Removed rows (since we don't track individual REMOVE operations in ps_oplog anymore).

crates/core/src/sync_local.rs
```diff
@@ -80,21 +114,34 @@ FROM updated_rows b
     LEFT OUTER JOIN ps_oplog AS r
                ON r.row_type = b.row_type
                  AND r.row_id = b.row_id
+                 AND r.bucket IN (SELECT id FROM involved_buckets)
```
rkistner, line 117: This bit has the potential to slow down the entire query quite a lot if there are many buckets.

crates/core/src/bucket_priority.rs
```diff
@@ -0,0 +1,89 @@
...
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct BucketPriority(i32);
+
+impl BucketPriority {
+    pub fn may_publish_with_outstanding_uploads(self) -> bool {
+        self == BucketPriority::HIGHEST
+    }
+
+    pub const HIGHEST: BucketPriority = BucketPriority(0);
+    pub const LOWEST: BucketPriority = BucketPriority(3);
```
rkistner, line 18: Can we remove the restriction on the lowest priority here? The only place I see it being used apart from validation is to detect the "sync everything" case for sync_local, which can also be done by just passing in a null parameter.

I'm thinking that we can leave it up to the service to specify the allowed range of priorities, and have 0 as the only special case on the client.

crates/core/src/checkpoint.rs
```diff
@@ -40,6 +40,8 @@ bucket_list(bucket, checksum) AS (
         json_extract(json_each.value, '$.bucket') as bucket,
         json_extract(json_each.value, '$.checksum') as checksum
   FROM json_each(json_extract(?1, '$.buckets'))
+  WHERE IFNULL(json_extract(json_each.value, '$.priority'), 1) <=
+    IFNULL(json_extract(?1, '$.priority'), 3)
```
rkistner, line 44: What is this check used for?

crates/core/src/sync_local.rs
rkistner: I added some individual comments on the query above, but I think overall we need a slightly different approach here. If specific buckets are specified, then `ps_updated_rows` should be completely ignored, since those entries are not grouped by bucket.

That does have the caveat that we cannot handle remove operations in partial checkpoints, but we just don't currently persist the required info to do that. I think the only way to solve that is to refactor ps_updated_rows to include bucket ids - not sure if we should do that now or in a next iteration.

For priority 0 - I'm not sure if we should care about local changes not uploaded yet. The idea for priority 0 is generally "sync anything from the server immediately". It removes all consistency guarantees, and the developer will generally need to use unique rows to avoid consistency issues (like with our Yjs demo where each update is persisted in a new row).

crates/core/src/checkpoint.rs
simolus3: The idea is that clients would send both the target priority and the priority of each bucket. Then we'd only validate the checksums of buckets matching that priority.

Clients can of course just send the buckets with a matching priority to begin with, which makes this check superfluous.

crates/core/src/sync_local.rs
simolus3: I've split the implementation to use different queries depending on whether we're doing a full or partial sync (that definitely needs some cleaning up, but that resolves this particular issue).

crates/core/src/sync_local.rs
rkistner: I think I understand why this is needed now.

One approach that _might_ be faster is to add another JOIN instead of the query:

```sql
INNER JOIN involved_buckets ON r.bucket = involved_buckets.id
```

I haven't tested that properly, but as far as I can see, the JOIN uses an automatic on the id of the materialized CTE, which the `WHERE ... IN` clause doesn't appear to use for some reason.

Either way, with the new approach of the split queries, the "normal" case without filtering buckets is not affected at least.

Also see my general notes on performance for this query in #56.

rkistner: For this specific bit of the query, I'm also trying to think on whether all the different edge cases are covered correctly, especially when the same rows are/were present in multiple buckets with different priorities.


Support we have two buckets: bucket1 and bucket2, with priorities 1 and 2 respectively.


<details>

<summary><h3>Hypothetical, if we tracked remove operations per bucket</h3></summary>

Case 1: A row is present in bucket 2, then added to bucket1. We get a partial_checkpoint_complete for bucket1.
The row will be included here, getting the latest version. :white_check_mark:	

Case 2: A row is removed from bucket2, and added to bucket1 at the same time (imagine "starring" a row to move it to a higher priority). We get a partial_checkpoint_complete for bucket1. The row will be included here, and whether or not we got the REMOVE on bucket2 already doesn't make a difference. When we sync the rest of the checkpoint, the updated row will stay present. :white_check_mark:	

Case 3: A row is removed from bucket1, and added to bucket2 at the same time (imaging removing a start to move to lower priority). We get a partial_checkpoint_complete for bucket1. The row will be removed here, whether or not we got the PUT for bucket2. This works "according to the spec", but not sure if this is desired behavior. :negative_squared_cross_mark:  When we sync the rest of the checkpoint, the row will be added again. :white_check_mark:

Case 4: A row is present in bucket1 and bucket2, then removed from bucket1. We get a partial_checkpoint_complete for bucket1. The row will be removed here, despite it still being present in bucket2. :negative_squared_cross_mark: When we sync the rest of the checkpoint, the row will _not_ be added again, resulting in an inconsistency. :x:

</details>
<details>

<summary><h3>Hypothetical, if we tracked remove operations per bucket</h3></summary>

Case 1: A row is present in bucket 2, then added to bucket1. We get a partial_checkpoint_complete for bucket1.
The row will be included here, getting the latest version. :white_check_mark:	

Case 2: A row is removed from bucket2, and added to bucket1 at the same time (imagine "starring" a row to move it to a higher priority). We get a partial_checkpoint_complete for bucket1. The row will be included here, and whether or not we got the REMOVE on bucket2 already doesn't make a difference. When we sync the rest of the checkpoint, the updated row will stay present. :white_check_mark:	

Case 3: A row is removed from bucket1, and added to bucket2 at the same time (imaging removing a start to move to lower priority). We get a partial_checkpoint_complete for bucket1. The row will be removed here, whether or not we got the PUT for bucket2. This works "according to the spec", but not sure if this is desired behavior. :negative_squared_cross_mark:  When we sync the rest of the checkpoint, the row will be added again. :white_check_mark:

Case 4: A row is present in bucket1 and bucket2, then removed from bucket1. We get a partial_checkpoint_complete for bucket1. The row will be removed here, despite it still being present in bucket2. :negative_squared_cross_mark: When we sync the rest of the checkpoint, the row will _not_ be added again, resulting in an inconsistency. :x:

</details>

rkistner:
After checking the implementation, I realized that a row present in multiple buckets with different priorities has a big potential for edge cases - both in the spec and the specific implementation. This is specifically relevant for the `r.bucket IN (SELECT id FROM involved_buckets)` clause for example.

A potential use case could be where a user has a _lot_ of data, and "stars" specific items to prioritize the sync of them.

For these examples, suppose we have two buckets: bucket1 and bucket2, with priorities 1 and 2 respectively. The same row could be in either or both of the buckets.

<details>

<summary><h3>Current state</h3></summary>

Case 1: A row is present in bucket 2, then added to bucket1. We get a partial_checkpoint_complete for bucket1.
The row will be included here, getting the latest version. :white_check_mark:

Case 2: A row is removed from bucket2, and added to bucket1 at the same time ("starring" a row to move it to a higher priority). We get a partial_checkpoint_complete for bucket1. The row will be included here, and whether or not we got the REMOVE on bucket2 already doesn't make a difference. When we sync the rest of the checkpoint, the updated row will stay present. :white_check_mark:

Case 3: A row is removed from bucket1, and added to bucket2 at the same time (removing a star to move to lower priority). We get a partial_checkpoint_complete for bucket1. We don't track removes per bucket, so this does nothing. :white_check_mark:  When we sync the rest of the checkpoint, the row will be added again. :white_check_mark:

Case 4: A row is present in bucket1 and bucket2, then removed from bucket1. We get a partial_checkpoint_complete for bucket1. We don't track removes per bucket, so this does nothing. :white_check_mark: When we sync the rest of the checkpoint, the row is updated with the state from bucket2. :white_check_mark:

</details>
<details>

<summary><h3>Hypothetical, if we tracked remove operations per bucket</h3></summary>

Case 1: A row is present in bucket 2, then added to bucket1. We get a partial_checkpoint_complete for bucket1.
The row will be included here, getting the latest version. :white_check_mark:	

Case 2: A row is removed from bucket2, and added to bucket1 at the same time ("starring" a row to move it to a higher priority). We get a partial_checkpoint_complete for bucket1. The row will be included here, and whether or not we got the REMOVE on bucket2 already doesn't make a difference. When we sync the rest of the checkpoint, the updated row will stay present. :white_check_mark:	

Case 3: A row is removed from bucket1, and added to bucket2 at the same time (removing a star to move to lower priority). We get a partial_checkpoint_complete for bucket1. The row will be removed here, whether or not we got the PUT for bucket2. This works "according to the spec", but not sure if this is desired behavior. :question:  When we sync the rest of the checkpoint, the row will be added again. :white_check_mark:

Case 4: A row is present in bucket1 and bucket2, then removed from bucket1. We get a partial_checkpoint_complete for bucket1. The row will be removed here, despite it still being present in bucket2. :question: When we sync the rest of the checkpoint, the row will _not_ be added again, resulting in an inconsistency. :x:

</details>

My summary from this that adding support for REMOVE operations in partial checkpoints does not actually give improved consistency as I hoped - it just creates more weird edge cases. The current behavior of only applying the REMOVE operations in the final checkpoint gives better results.

crates/core/src/sync_local.rs
rkistner: Some notes for myself: I thought it may be feasible to remove the `r.bucket IN (SELECT id FROM involved_buckets)` clause, since the higher priority buckets should generally have "newer" data, which means the non-involved bucket data would not take precedence either way. However, I think there are edge cases where the assumption breaks, so it's a bad idea to rely on that.

Specific contrived example: Download a row in bucket1b, then interrupt the connection before sync_local. By the time the sync restarts, bucket1b was removed, but we sync a row with a lower op_id in bucket1a (this bit is contrived, but technically possible). Now, sync_local with bucket1a would persist the downloaded copy from bucket1b, which should never have happened. Filtering by "involved_buckets" prevents this from happening.

simolus3: > One approach that _might_ be faster is to add another JOIN instead of the query:

I didn't check, but I would kind of expect the query planer to figure this one out :) I can change it to a join nonetheless, but I'm also not sure if we need it now that the main query for full syncs is no longer affected.

rkistner: I'm happy to keep as-is for this release. I'll re-investigate and get some proper testing up as part of #56.

crates/core/src/sync_local.rs
```diff
@@ -288,19 +291,35 @@ GROUP BY b.row_type, b.row_id",
...
-            // language=SQLite
-            self.db
-                .exec_safe(
-                    "insert or replace into ps_kv(key, value) values('last_synced_at', datetime())",
-                )
-                .into_db_result(self.db)?;
-        }
+        // Higher-priority buckets are always part of lower-priority sync operations too, so we can
+        // delete information about higher-priority syncs (represented as lower priority numbers).
+        // A complete sync is represented as -1.
```
rkistner, line 307: The `-1` here makes the sort ordering a bit weird. You essentially have:

```
0: synced priority 0
1: synced priority 0 + 1
2: synced priority 0 + 1 + 2
3: synced priority 0 + 1 + 2 + 3
-1: synced priority 0 + 1 + 2 + 3
```

An example if this having in effect is in the delete query below (`(priority < ?1) OR (?1 = -1)`).

Could we perhaps use a constant for this that is actually bigger than the other priority values?

simolus3: I've added `i32::MAX` as a priority representing a complete sync.

rkistner (approved):

PR #66, opened 2025-03-20
Add table option to skip empty updates

simolus3:
For updates on views that happen to not change any columns (e.g. a `SET col = LOWER(col)` where all values in `col` were already in lower case), we currently insert an empty update operation into the local upload queue. While this can be resolved by users rewriting SQL statements and typically shouldn't make up a bulk of the queue, we've seen a report claiming that skipping empty updates speeds up uploads: https://github.com/powersync-ja/powersync-kotlin/pull/144.

Ignoring empty updates entirely might break users relying on the current behavior, so this adds an opt-in option to ignore schema updates. For this, the `powersync_crud_` vtab gains a new hidden column with option flags that can be set for inserts. `FLAG_IGNORE_EMPTY_UPDATE`, which is disabled by default, would make an insert into `ps_crud` do nothing if it's an empty update.


crates/core/src/crud_vtab.rs
```diff
@@ -69,15 +76,22 @@ extern "C" fn disconnect(vtab: *mut sqlite::vtab) -> c_int {
     ResultCode::OK as c_int
 }
 
-
 fn begin_impl(tab: &mut VirtualTable) -> Result<(), SQLiteError> {
     let db = tab.db;
 
-    let insert_statement = db.prepare_v3("INSERT INTO ps_crud(tx_id, data) VALUES (?1, ?2)", 0)?;
+    const SQL: &str = formatcp!("\
+WITH insertion (tx_id, data) AS (VALUES (?1, ?2))
```
rkistner, line 83: To check if there is data in the patch, it might be more efficient to just check `data->'data' != '{}'` - `json_each` could potentially add some overhead here. I haven't done any actual benchmarks though - the overhead may be negligible.

rkistner (commented):
Happy with the approach

rkistner (approved):

PR #67, opened 2025-03-25
Fix migration tests

rkistner:
Fix test fixtures after #65.

This now also configures the tests to be able to run on pull requests from external contributors (after approval), to catch these type of issues earlier. The `if` in the tests is so that a pull request by a core contributor doesn't trigger the workflow via both `push:` and `pull_request:` runs - that becomes a lot.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -1,9 +1,11 @@
 on:
   push:
+  pull_request:
 name: "android"
 jobs:
   build:
     name: Building Android
+    if: github.event_name == 'push' || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository)
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -1,9 +1,11 @@
 on:
   push:
+  pull_request:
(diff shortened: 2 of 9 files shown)


simolus3 (approved):

.github/workflows/android.yml
```diff
@@ -1,9 +1,11 @@
 on:
   push:
+  pull_request:
```
simolus3, line 3: For the Kotlin SDK we're doing [this](https://github.com/powersync-ja/powersync-kotlin/blob/1d939cc76e9501a5d81dfb8d7c843ba2f66e46ac/.github/workflows/test.yml#L3-L8) which may be nice to avoid the "skipped" list of actions on every PR we open now:

```YAML
on:
  push:
    branches: [ "main" ]
  pull_request:
    branches: [ "main" ]
  workflow_call:
```

It would have the downside that our branches only get tested after we open a PR, but that migh be fine.

rkistner: I'm annoyed that github doesn't have something better built-in for this. I also used the workaround here in [powersync-service](https://github.com/powersync-ja/powersync-service/pull/205), and it seems to work quite well: Same behavior as currently (build on each push) for branches within the repo, and switches to the "pull_request" builds when it's from a fork. Main downside is the extra boilerplate in the workflows.

PR #69, opened 2025-04-09
Expand support for update metadata and tracking old values

simolus3:
This is a follow-up for https://github.com/powersync-ja/powersync-sqlite-core/pull/60 to:

- Support tracking old values with `include_old` for `DELETE` statements.
- Support storing old values only if they have changed (through `include_old_only_when_changed`).
- Support storing old values only for a subset of column names (by setting `include_old` to a JSON array string).
- Support `include_metadata` for inserts.
- Support `include_metadata` for deletes with a special `UPDATE <table> SET _deleted = TRUE` syntax (we can't attach data to the delete otherwise).

It also refactors some of the implementation details, in particular:

1. This updates the `sqlite-rs` version, and then applies our fork that makes `ResultCode` implement `Clone` and `Copy`.
2. Refactors extracting table options into a typed `TableInfo` struct.
3. Moves the logic for creating and iterating over the `columns` array of a schema table into a shared struct utilizing the `streaming_iterator` create. This makes it easier to pass custom iterators to `json_object_fragment`, which is used to restrict `include_old` to a list of columns.
4. Simplifies the `format!` logic to create triggers by using inline references, making it easier to understand. (Eventually, I want to spend some time eventually to that further by e.g. using `write!`, we're using lots of small allocations and duplicate logic here.)


crates/core/src/schema/table_info.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    pub const fn insert_only(self) -> bool {
+        self.0 & Self::INSERT_ONLY != 0
+    }
+
+    pub const fn include_metadata(self) -> bool {
+        self.0 & Self::INCLUDE_METADATA != 0
+    }
+
+    pub const fn include_old_only_if_changed(self) -> bool {
```
rkistner, line 121: Can we make this `include_old_only_when_changed` to make it consistent with the json option passed in?

crates/core/src/views.rs
```diff
@@ -206,76 +260,98 @@ fn powersync_trigger_update_sql_impl(
...
+                    Some(Ok(iterator.next()?.as_str()))
+                });
+
+            Some(json_object_fragment("OLD", &mut columns)?)
+        }
+    };
+
+    if table_info.flags.include_old_only_if_changed() {
+        // We're setting ignore_removed_columns here because values that are present in the new
+        // values but not in the old values should not lead to a null entry.
```
rkistner, line 296: Can you explain the reasoning here - why it should not lead to a null entry?

It would also be good to have a unit test for this specific case, demonstrating the behavior.

From my own testing, this appears to specifically be applicable to a case where a value is changed from null to another value. With `ignore_removed_columns`, `old.<column>` would not be present. Without it, it would be null. Either way you know it has changed due to the column being present in the "new" values. So I think it should be fine either way, but I'm still wondering on the specific reasoning here.

crates/core/src/diff.rs
```diff
@@ -20,11 +19,20 @@ fn powersync_diff_impl(
...
 
-    diff_objects(data_old, data_new)
+    diff_objects_with_options(data_old, data_new, ignore_removed)
 }
 
-pub fn diff_objects(data_old: &str, data_new: &str) -> Result<String, SQLiteError> {
+/// Returns a JSON object containing entries from [data_new] that are not present in [data_old].
+///
+/// When [ignore_removed_columns] is set, columns that are present in [data_old] but not in
+/// [data_new] will not be present in the returned object. Otherwise, they will be set to `null`.
```
rkistner, line 30: It may be useful to describe the specific use case here (calculating the "old" values with `include_old_only_when_changed`).

crates/core/src/views.rs
simolus3: I think I ran into this writing the test `crud tests tracks previous values for updates combined column filter and only tracking changes`. The problem is that we're passing `json_fragment_new` to `powersync_diff` - which includes all columns. The `old_values_fragment` however, does not. So there was a possibility where unchanged columns would be reported just because they're absent in `old_values_fragment` as they're not matched by the filter.

I don't remember why I came up with that workaround - a better solution is to just apply the same filter to the fragment producing the new columns too. That's what I've done now.

rkistner (approved):

rkistner (approved):

PR #70, opened 2025-04-23
Implement sync client

simolus3:
This implements the sync client logic in the core extension. Client SDKs will still be responsible for opening sockets to the sync service, but the core extension is now responsible for driving that logic. The central interface here is the `powersync_control` SQL function, which is invoked by clients for a request to start a sync stream or lines received by the sync service. In response, the core extension responds with a list of instruction for the SDK (like e.g. updating the sync status).

We expect that implementing the stream client here will:

1. Improve performance around handling BSON lines: Only our JavaScript SDK supports BSON at the moment. At the moment, that SDK has to deserialize the BSON line itself and re-encode it as JSON before sending it to the core extension. By implementing the deserialization in the core extension, we avoid that inefficiency.
2. Make the protocol easier to evolve: Subsequent changes would mostly have to implemented here instead of at each client SDK.
3. Make it easier to adopt the RSocket protocol across our SDKs.
4. Not done here: Eventually support the BLOB column type, which requires a BSON-based transport that can now be handled here for all our SDKs.

The most important are structured like this:

- `core/src/bson` contains a no-std implementation of the BSON format with serde support. The `bson` crate unfortunately only supports std environments and I couldn't find a decent alternative. We don't support all BSON types, but that should be fine since we only need the types supported by JSON (as well as byte arrays).
- `core/src/sync` contains logic to implement the sync client:
  - `line.rs` defines the sync lines as structs supporting deserialization.
  - `storage_adapter.rs` is used as an interface for some common database operations. It's a bit simpler than the `BucketStorage` interface in the SDK since we can mostly just call other core functions directly without going through SQL.
  - `sync_status.rs` implements the `SyncStatus` interface from our other SDKs and logic to update it along with sending update notifications to the client.
  - `interface.rs` defines the `powersync_control` SQL function driving the sync client.
  - `streaming_sync.rs` contains the actual state machine for the sync client. I've implemented it as an async coroutine that is polled every time `powersync_control` is called, which allows an implementation that looks similar to the ones across our other SDKs.

### DIFF
--- a/crates/core/src/sync/line.rs
+++ b/crates/core/src/sync/line.rs
@@ -2,16 +2,36 @@ use alloc::borrow::Cow;
 use alloc::vec::Vec;
 use serde::Deserialize;
 
-use super::BucketPriority;
-use super::Checksum;
-
 use crate::util::{deserialize_optional_string_to_i64, deserialize_string_to_i64};
+
+use super::bucket_priority::BucketPriority;
+use super::Checksum;
 
 /// While we would like to always borrow strings for efficiency, that's not consistently possible.
 /// With the JSON decoder, borrowing from input data is only possible when the string contains no
 /// escape sequences (otherwise, the string is not a direct view of input data and we need an
 /// internal copy).
 type SyncLineStr<'a> = Cow<'a, str>;
+
+#[derive(Deserialize, Debug)]
+
+pub enum SyncLine<'a> {
+    #[serde(rename = "checkpoint", borrow)]
+    Checkpoint(Checkpoint<'a>),
+    #[serde(rename = "checkpoint_diff", borrow)]
+    CheckpointDiff(CheckpointDiff<'a>),
+
+    #[serde(rename = "checkpoint_complete")]
+    CheckpointComplete(CheckpointComplete),
+    #[serde(rename = "partial_checkpoint_complete")]
+    CheckpointPartiallyComplete(CheckpointPartiallyComplete),
+
+    #[serde(rename = "data", borrow)]
+    Data(DataLine<'a>),
+
+    #[serde(rename = "token_expires_in")]
+    KeepAlive(TokenExpiresIn),
+}
 
 #[derive(Deserialize, Debug)]
 pub struct Checkpoint<'a> {
@@ -22,6 +42,32 @@ pub struct Checkpoint<'a> {
     pub write_checkpoint: Option<i64>,
     #[serde(borrow)]
     pub buckets: Vec<BucketChecksum<'a>>,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct CheckpointDiff<'a> {
+    #[serde(deserialize_with = "deserialize_string_to_i64")]
+    pub last_op_id: i64,
+    #[serde(borrow)]
+    pub updated_buckets: Vec<BucketChecksum<'a>>,
+    #[serde(borrow)]
+    pub removed_buckets: Vec<SyncLineStr<'a>>,
+    #[serde(default)]
+    #[serde(deserialize_with = "deserialize_optional_string_to_i64")]
+    pub write_checkpoint: Option<i64>,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct CheckpointComplete {
+    //    #[serde(deserialize_with = "deserialize_string_to_i64")]
+    //    pub last_op_id: i64,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct CheckpointPartiallyComplete {
+    //    #[serde(deserialize_with = "deserialize_string_to_i64")]
+    //    pub last_op_id: i64,
+    pub priority: BucketPriority,
 }
 
 #[derive(Deserialize, Debug)]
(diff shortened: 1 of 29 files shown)


crates/core/src/json.rs
```diff
@@ -0,0 +1,281 @@
...
+    powersync_json_merge,
+    powersync_json_merge_impl,
+    "powersync_json_merge"
+);
+
+/// A variant of `json_extract` that works both with JSON and BSON.
+///
+/// We only support extracting top-level keys here, so an invocation would look like
+/// `powersync_extract('{"foo": "bar"}', 'foo')`.
+fn powersync_extract_impl(
```
rkistner, line 60: Just an early comment here: If you have a typical PowerSync table with say 50 columns, that would mean 50x `powersync_extract` calls for each row in the view, which could mean re-parsing the same JSON/BSON 50 times.

I believe the built-in SQLite json support uses some form of caching of parsed json, which could have a significant impact here. I think it relies on some undocumented behavior though.

I've never tested the performance of this, but it could have quite a big impact for tables with large numbers of columns.

crates/core/src/views.rs
```diff
@@ -19,35 +22,43 @@ fn powersync_view_sql_impl(
...
-    while stmt2.step()? == ResultCode::ROW {
-        let name = stmt2.column_text(0)?;
-        let type_name = stmt2.column_text(1)?;
+    while let Some(row) = iter.next() {
+        let ColumnInfo { name, type_name } = row.clone()?;
         column_names_quoted.push(quote_identifier(name));
 
         let foo = format!(
-            "CAST(json_extract(data, {:}) as {:})",
+            "CAST(powersync_extract(data, {:}) as {:})",
```
rkistner, line 47: An advantage of the old `json_extract` form was that you could read a powersync db without loading the extension. Is there a nice way to keep that option available, for example when binary data isn't used?

crates/core/src/sync/line.rs
```diff
@@ -0,0 +1,450 @@
...
+
+                Ok(OplogData::JsonString {
+                    data: Cow::Owned(writer.finish()),
+                })
+            }
+        }
+
+        // Regardless of whether we're deserializing JSON or BSON, oplog data is represented either
+        // as a string (representing a JSON-encoded object) or an object (representing the values
+        // directly).
```
rkistner, line 270: In the current protocol version, the data is _always_ a string (JSON-encoded object), both for JSON and BSON encoding of the metadata. It seems like the code here can be reduced quite a bit by only catering for that case.

## Some background:

For backwards-compatibility with old (ancient) clients, there is an option for the server to respond with embedded JSON objects instead, but that adds a lot of overhead on the service due to using a [slower serializer](https://github.com/powersync-ja/powersync-service/blob/5b390393f3d03fc5422fbb00c1c5a4cd74b15c3a/packages/service-core/src/sync/sync.ts#L384-L387) for that case. We haven't used this option in any SDK in a long time.

In general, we switching to keeping the data as a string everywhere, to avoid having to parse + serialize json objects repeatedly as it is being passed from bucket storage -> NodeJS -> client -> sqlite. Now, there is still parser overhead when using NDJSON, due to having to encode the string in the JSON, but at least in memory it stays a plain string which has a big performance impact in some cases.

Now I realize that in this specific implementation, you can actually optimize the raw JSON version by keeping the embedded object as-is. That is not an optimization we had available in most SDKs with generic JSON parsers. But I think it's still better to stick to just the string version, due to the reduced overhead on the service. And switching to BSON will remove that overhead either way.

The same applies to the BSON encoding - when we do support BSON data, we'd likely keep the embedded data as a "binary" field, rather than an embedded document, so that that the data can just pass it through as-is even when using more high-level parsers/serializers.

rkistner (commented):
I'm reading through the process flow and associated state management now, and just want to check that I understand it correctly:
1. A `SqlController` with `SyncClient` is created per connection when the extension is registered. This keeps a `ClientState`, which is `Idle`/empty, until "start" even is received.
2. When the "start" even is received, it initializes a `SyncIterationHandle`, stored n the `ClientState`, and containing a `StreamingSyncIteration`. A mutex on the `ClientState` protects against instantiating multiple copies of this at the same time. (Is this correct? Not sure if I understand the purpose of the "Mutex" here correctly). Once a "stop" event is received, this is disposed.
3. The `StreamingSyncIteration` contains a `StorageAdapter`, which includes some prepared statements. These are all disposed automatically when the `StreamingSyncIteration` is disposed via the "stop" event.

It  seems like this gives us a nice way to persist prepared statements, instead of re-creating them every time. I've never measured the performance overhead, but I imagine there are places where persisting the prepared statements could reduce the performance overhead quite a bit. One disadvantage being that we'd still need to cater for the older APIs, which don't have a nice place to persist those statements. Either way, that is not important for the initial iteration here - we can do those optimizations later.

crates/core/src/sync/line.rs
simolus3: That's good to know, and only supporting JSON strings here (with the intention or later supporting BSON data too) definitely simplifies this quite a bit. I've also opened a PR on the sync service to clarify the types there, I was under the impression that we have to support those based on the way `OplopEntry` is defined there.

docs/sync.md
```diff
@@ -0,0 +1,79 @@
...
+
+1. `start`: Payload is a JSON-encoded object. This requests the client to start a sync iteration, specifying
+   parameters.
+2. `stop`: No payload, requests the current sync iteration (if any) to be shut down.
+3. `line_text`: Payload is a serialized JSON object received from the sync service.
+4. `line_binary`: Payload is a BSON-encoded object received from the sync service.
+5. `refreshed_token`: Notify the sync client that the JWT used to authenticate to the PowerSync service has
+   changed.
+   - The client will emit an instruction to stop the current stream, clients should restart by sending another `start`
+     command.
```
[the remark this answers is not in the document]
rkistner: Happy with this, just wanted to confirm.

kobiebotha (approved):

PR #73, opened 2025-05-11
feat: add watchOS support (tentative)

zifeo:
Not a watchOS expert, so carefully review and guidance required.

### DIFF
--- a/tool/build_xcframework.sh
+++ b/tool/build_xcframework.sh
@@ -3,12 +3,11 @@ set -e
 
 # Adapted from https://github.com/vlcn-io/cr-sqlite/blob/main/core/all-ios-loadable.sh
 
-
 BUILD_DIR=./build
-DIST_PACKAGE_DIR=./dist
 
 function createXcframework() {
-  plist=$(cat << EOF
+  ios_plist=$(
+    cat <<EOF
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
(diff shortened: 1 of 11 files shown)


tool/build_xcframework.sh
```diff
@@ -3,12 +3,11 @@ set -e
 
 # Adapted from https://github.com/vlcn-io/cr-sqlite/blob/main/core/all-ios-loadable.sh
 
-
 BUILD_DIR=./build
-DIST_PACKAGE_DIR=./dist
 
 function createXcframework() {
-  plist=$(cat << EOF
+  ios_plist=$(
```
simolus3, line 9: It's technically not just ios, we're using this for macOS too. I don't know what a better name would be either though, maybe `common_plist`?

simolus3 (commented):
Thank you for getting started on this!

I've started looking into support for this for our Kotlin SDK here: https://github.com/powersync-ja/powersync-kotlin/tree/watchOS

tool/build_xcframework.sh
```diff
@@ -62,20 +111,37 @@ EOF
...
     -framework "${BUILD_DIR}/macos-arm64_x86_64/powersync-sqlite-core.framework" \
     -debug-symbols "$(pwd -P)/${BUILD_DIR}/macos-arm64_x86_64/powersync-sqlite-core.framework.dSYM" \
-    -output "${BUILD_DIR}/powersync-sqlite-core.xcframework" \
+    -output "${BUILD_DIR}/powersync-sqlite-core.xcframework"
+
+  # how to create a watchOS XCFramework with static libraries, possible?
 
-  cp -Rf "${BUILD_DIR}/powersync-sqlite-core.xcframework" "powersync-sqlite-core.xcframework"
-  zip -r --symlinks powersync-sqlite-core.xcframework.zip powersync-sqlite-core.xcframework LICENSE README.md
+  zip -r --symlinks powersync-sqlite-core.xcframework.zip powersync-sqlite-core.xcframework "${BUILD_DIR}/watchos-arm64/powersync-sqlite-core.framework/libpowersync.a" "${BUILD_DIR}/watchos-arm64-simulator/powersync-sqlite-core.framework/libpowersync.a" LICENSE README.md
```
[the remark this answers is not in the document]
simolus3: For context: Rustc refuses to build dynamic libraries for watchOS. So yes, this is required.

crates/static/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+version.workspace = true
+homepage.workspace = true
+repository.workspace = true
+license.workspace = true
+authors.workspace = true
+keywords.workspace = true
+
+[lib]
+name = "powersync"
+crate-type = ["staticlib"]
```
simolus3, line 13: I've added this as a separate package because building `powersync_loadable` as a dynamic library is not supported for watchOS builds (and `cargo build -p powersync_loadable` will build both a static and a dynamic library, I couldn't find a CLI flag to disable the dynamic build).

So I've created this new crate for static builds that we're now using for WASM and watchOS.

Package.swift
```diff
@@ -0,0 +1,26 @@
+// swift-tools-version: 5.7
+
+// NOTE! This is never released, we're only using this to support local builds builds for the
```
simolus3, line 3: I've added this package so that we can use a local dependency from the Swift SDK to this directory. This allows testing new versions of the core extension with the Swift SDK, which was was quite cumbersome before.

simolus3 (commented):
I've tested this against the Kotlin and Swift SDKs, so this looks good to me.

rkistner (approved):

zifeo:
Thanks for pushing this further, I lack time to take care of it in the past days!

PR #79, opened 2025-05-27
sync_local: Re-use prepared statements

rkistner:
This keeps prepared statements around for the last table, to avoid preparing a new statement for every row that we insert/delete.

I did not fully confirm, but I expect that the way the query plan is structured that the rows would typically be processed one table at a time, making it efficient to only keep the tables for the last query. It is not an issue if the order is different - we may just lose a little of the performance benefits here.

In my tests on native desktop with an in-memory db, this reduced the overall sync_local duration by around 25% (2786ms -> 2075ms for 500k rows). When taking into account filesystem overhead, the impact will be a little less, but still significant.

### DIFF
--- a/crates/core/src/sync_local.rs
+++ b/crates/core/src/sync_local.rs
@@ -103,7 +103,17 @@ impl<'a> SyncOperation<'a> {
 
         self.collect_tables()?;
         let statement = self.collect_full_operations()?;
-        // TODO: cache statements
+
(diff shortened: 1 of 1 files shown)


simolus3 (dismissed):
Looks good to me 👍 I agree that we don't have to worry about enforcing table order here.

simolus3 (approved):

PR #80, opened 2025-05-30
Add checksum newtype

simolus3:
I'm moving some independent parts out of #70 to make that easier to review and understand.

This adds a `Checksum` newtype wrapping a `u32`, making the way we deal with checksum computations a little easier to understand:

1. `Checksum` implements the `+` and `-` operators modulo 2³². This is the behavior we want, and avoids issues if we forget a `wrapping_add` or `wrapping_sub` call somewhere.
2. It centralizes the fact that checksums are `u32` types. We've previously been using `i32` (which doesn't make a difference for addition and subtraction), but this makes it clearer that checksums are unsigned types that need to be explicitly cast from `i32` where necessary.
3. This moves some sync-related files into a `sync` submodule.


rkistner (dismissed):
I like this approach.

Could we also add one or two tests for the wrapping add/subtract behavior?

rkistner (approved):

PR #85, opened 2025-06-10
Add BSON deserialize tests

rkistner:
Add some BSON parsing tests, to make sure we parse all the types correctly. Even with the official BSON libraries I've seen some edge cases handled incorrectly ([example](https://github.com/powersync-ja/powersync-service/pull/207)), so I feel testing is important.

These tests cases were automatically generated by claude code, but I manually confirmed that the output matches that of the JS bson package.


simolus3 (approved):
Nice, thanks 👍 I didn't check all the input data either, but the test cases themselves make sense to me and since we match the behavior of another BSON library, this looks good to me.

PR #87, opened 2025-06-11
Raw tables

simolus3:
This adds support for raw tables which, instead of being views over JSON data, are meant to be created by users as regular tables.

This prepares the core extension with support for that, but there are still a lot of uncertainties around the exact interface to use. As an experiment here, we:

1. Expand the schema JSON to include an additional `raw_tables` array in additon to the array of JSON-backed tables. Raw tables are not created by the core extension.
2. Each raw table has a name identifiying it, as well as an SQL statements for upserts (to handle `PUT` and `PATCH` operations) and deletes.
3. These statements can use parameters, which the core extension can bind to column values (either the ID or extracted data).
4. We need to be aware of the schema during the sync operation to know which statements to use. This feature relies on the new sync logic in the core extension, and involves clients passing the schema to the first `powersync_control` invocation.

For local writes, clients would typically install triggers on raw tables. These triggers need to be able to distinguish between actual local writes from the app and writes made to apply sync operations. For that, this:

1. Introduces the `powersync_in_sync_operation()` function which returns `TRUE` when a sync is active on the connection (so `WHEN NOT powersync_in_sync_operation()` can be used as a guard on the trigger).
2. Also changes `ps_crud` vtab insertions to be ignored during a local sync operation.

### DIFF
--- a/crates/core/src/sync_local.rs
+++ b/crates/core/src/sync_local.rs
@@ -501,8 +501,20 @@ impl<'a> PreparedPendingStatement<'a> {
     pub fn prepare(
         db: *mut sqlite::sqlite3,
         pending: &'a PendingStatement,
-    ) -> Result<Self, ResultCode> {
+    ) -> Result<Self, SQLiteError> {
         let stmt = db.prepare_v2(&pending.sql)?;
+        if stmt.bind_parameter_count() as usize != pending.params.len() {
+            return Err(SQLiteError(
+                ResultCode::MISUSE,
+                Some(format!(
+                    "Statement {} has {} parameters, but {} values were provided as sources.",
+                    &pending.sql,
+                    stmt.bind_parameter_count(),
+                    pending.params.len(),
+                )),
+            ));
+        }
+
         // TODO: Compare number of variables / other validity checks?
 
         Ok(Self {
@@ -534,8 +546,7 @@ impl<'a> PreparedPendingStatement<'a> {
                         }
                         Some(Value::Number(value)) => {
                             if let Some(value) = value.as_f64() {
-                                // ??? there's no bind_double???
-                                self.stmt.bind_int64(i, value as i64)
+                                self.stmt.bind_double(i, value)
                             } else if let Some(value) = value.as_u64() {
                                 self.stmt.bind_int64(i, value as i64)
                             } else {


crates/core/src/sync_local.rs
```diff
@@ -372,3 +398,175 @@ SELECT
...
+                    })?;
+
+                    match parsed.get(column) {
+                        Some(Value::Bool(value)) => {
+                            self.stmt.bind_int(i, if *value { 1 } else { 0 })
+                        }
+                        Some(Value::Number(value)) => {
+                            if let Some(value) = value.as_f64() {
+                                // ??? there's no bind_double???
+                                self.stmt.bind_int64(i, value as i64)
```
rkistner, line 538: What will happen in this case - I assume the value is rounded down? Should we rather error or use null in this case for now, until we add support in `sqlite-rs-embedded`?

rkistner (commented):
Overall I'm quite happy with this.

I wonder if we should also support a simplified syntax for raw tables using named parameters, without requiring the separate parameters array? I'm thinking a syntax like this:

```sql
INSERT OR REPLACE INTO users (id, name) VALUES (:id, :name)
```

In theory it may be possible to combine the two, but parameter numbering becomes a little complex. But either way, we don't need this from the start, but it could be useful to have the "shorthand" syntax in the future.

crates/core/src/sync_local.rs
simolus3: Let's just do it right from the beginning: https://github.com/powersync-ja/sqlite-rs-embedded/pull/3

simolus3:
> In theory it may be possible to combine the two, but parameter numbering becomes a little complex. But either way, we don't need this from the start, but it could be useful to have the "shorthand" syntax in the future.

I agree 👍 One reason for the current JSON-based parameter sources is that we'd be able to expand them in the future, e.g. we could add a "`PUT` data as JSON string" source and let users do the extraction on their own. It would be hard to find names for this that don't clash with column names, but I think we could still do it with naming conventions like e.g. `:columnName` vs. `$jsonSource`.

I also think that doing it later after we know we have all the functionality we need for this could be more helpful - we're not concerned with breaking changes here since it's an experimental feature.

PR #91, opened 2025-06-23
Crud vtab optimizations

simolus3:
This optimizes the crud vtab to:

1. Instead of preparing statements on `BEGIN`, prepare them lazily on the first write to the crud vtab in a transaction. This reduces the overhead of that virtual table in unrelated transactions.
2. In the simple mode, only execute the statement to update `$local` once.

### DIFF
--- a/crates/core/src/crud_vtab.rs
+++ b/crates/core/src/crud_vtab.rs
@@ -160,7 +160,7 @@ impl VirtualTable {
                 set_updated_rows.bind_text(1, row_type, sqlite::Destructor::STATIC)?;
                 set_updated_rows.bind_text(2, id, sqlite::Destructor::STATIC)?;
                 set_updated_rows.exec()?;
-                simple.had_writes = true;
+                simple.record_local_write(db)?;
             }
         }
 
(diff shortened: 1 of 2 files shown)


crates/core/src/crud_vtab.rs
```diff
@@ -176,41 +182,90 @@ impl VirtualTable {
...
+            let needs_local_bucket_update = match tx.mode {
+                CrudTransactionMode::Manual { .. } => {
+                    // In manual mode, users need to update the $local bucket themselves.
+                    false
+                }
+                CrudTransactionMode::Simple(simple) => simple.had_writes,
+            };
+
+            if needs_local_bucket_update {
+                self.db.exec_safe(formatcp!("INSERT OR REPLACE INTO ps_buckets(name, last_op, target_op) VALUES('$local', 0, {MAX_OP_ID})"))?;
```
rkistner, line 206: Is it safe to do this inside the `xCommit` callback? As in, are we sure that this will be called before the sync/commit process is started on the underlying VFS? I can't find anything mentioning the order of operations in the [vtab docs](https://www.sqlite.org/vtab.html), so not sure how to check that.

simolus3: Good catch! I think the `xSync` method would be a better fit for this, and `xCommit` definitely isn't safe. However, we can also just run the statement once on the first write which sounds like the easiest option.

rkistner (approved):

PR #108, opened 2025-07-11
Avoid nested JSON elements in ps_crud

simolus3:
To collect local writes, we use `json_object()` to combine column values into a CRUD entry. A problem with this is that, if one of the columns has a JSON subtype, `json_object()` writes it as a nested object instead of a string. Following [this suggestion](https://www.sqlite.org/forum/forumpost/57ed0b1feb30bd01), we wrap text column types in a single-argument `concat` call to drop the subtype.

I have not yet checked the performance impact of this. Another approach that unfortunately occurred too late to me may be to re-implement `json_object` in PowerSync, avoiding the nested behavior we don't need. That could potentially be faster.

### DIFF
--- a/crates/core/src/schema/mod.rs
+++ b/crates/core/src/schema/mod.rs
@@ -6,7 +6,8 @@ use serde::Deserialize;
 use sqlite::ResultCode;
 use sqlite_nostd as sqlite;
 pub use table_info::{
-    DiffIncludeOld, PendingStatement, PendingStatementValue, RawTable, Table, TableInfoFlags,
+    Column, DiffIncludeOld, PendingStatement, PendingStatementValue, RawTable, Table,
+    TableInfoFlags,
 };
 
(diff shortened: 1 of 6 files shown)


rkistner (approved):
I did some very basic performance tests, comparing the performance of these two queries:

```sql
select sum(length(json_object('t', data))), count() from ps_oplog;
select sum(length(json_object('t', concat(data)))), count() from ps_oplog;
```

In these cases, `data` is always a string, so I'm just measuring the overhead when no actual conversion is required.

In my test set, I have 114k rows, totaling around 100MiB in size. The first version takes an average of 117ms, while the `concat` version takes an average of 120ms. And this is a large data set, much larger than typical data we'd see in the upload queue. Compared to the rest of the overhead we have in the upload queue, including actually writing the data, this overhead appears to be negligible.

PR #121, opened 2025-08-13
Refactor CI build

simolus3:
This refactors the CI build with the goal of making non-released branches easier to test and to re-use build logic between the test and the release workflow.

Similar to the setup we already have for Android, this moves the build of all binaries into a composite `.github/actions/` action. The last step of that action is to upload the binaries it compiled as an asset. For Dart tests and the release, those assets are downloaded instead of compiling the core extension again. A neat side-effect of this is that we're guaranteed to test the libraries in the exact same way as we'll release them. Since the retention time of artifacts has been increased to 14 days, it has become a bit easier to test unreleased branches by updating download URLs in client packages (previously, one would always have to use a local build of the core extension).

I've also changed the linux build to run on x86-64 only, cross-compiling for multiple targets instead of using an aarch64 host. That also adds builds for armv7 and riscv64gc Linux, ensuring we support all Linux targets supported by the Dart SDK.
In addition to the x86-64 Linux and aarch64 macOS targets previously used to run Dart tests, we're now running Dart tests on aarch64 ubuntu and both x86-64 and aarch64 Windows too.


simolus3:
That list is correct for Windows, Linux and Android. Since recently, the xcframework contains additional targets (aarch64-apple-watchos, aarch64-apple-watchos-sim, x86_64-apple-watchos-sim, arm64_32-apple-watchos, aarch64-apple-tvos, aarch64-apple-tvos-sim, x86_64-apple-tvos). It's also worth noting that GH is in the process of upgrading `macos-latest` to macOS 15, I'm not sure if that will also include a switch to aarch64 being the default. But it shouldn't matter whether we're cross-compiling from x64 or aarch64.

rkistner:
>  Since recently, the xcframework contains additional targets

I see, I updated the list. I got the initial list from the `rustup target add`, and missed that the actual build script has additional targets. The Rust build chains gets a little confusing between the toolchain & rust-src component, targets added, and targets finally built - not sure if the additional targets are auto-installed, and if we need the initial `rustup target add` at all?

PR #128, opened 2025-09-09
Reduce warnings for sync after uploads

simolus3:
Consider the following sequence of events:

1. The client is offline and mutates the database.
2. The client connects, and starts uploading changes.
3. That triggers a new checkpoint that gets synced down. It can't be applied because `$local.target_op` is at  1<<62.
4. The connector is done, the client SDK requests a write checkpoint.
4. We get a write checkpoint response, it's `1`.
6. Because the upload has finished, we send an `upload_completed` notification to the Rust client. That tries applying the previous checkpoint, which fails because it doesn't have a write checkpoint. A "Could not apply pending checkpoint even after completed upload" warning message is logged.
7. A new checkpoint with the write checkpoint is synced down and applied.

This is actually fairly common, and a completely harmless sequence of events that both the service and the client handle correctly. It's just the warning message that stands out here. So, this PR removes that message.

This also adds tests for:

1. The checkpoint containing a requested write checkpoint being synced before the crud upload task in SDKs completes (completing it should apply the checkpoint, recently handled correctly).
2. The crud upload completing and triggering a new checkpoint with the data and write checkpoint (we've always handled this one correctly).
3. Receiving a data checkpoint first and a write checkpoint later (there shouldn't be the annoying warning message).
4. A second local write after the first one has been uploaded (this should still trigger the warning message to notify developers about why the sync couldn't be completed).

Closes https://github.com/powersync-ja/powersync-kotlin/issues/243.

### DIFF
--- a/crates/core/src/sync/streaming_sync.rs
+++ b/crates/core/src/sync/streaming_sync.rs
@@ -448,6 +448,12 @@ impl StreamingSyncIteration {
                     |s| s.start_tracking_checkpoint(progress, subscription_state),
                     &mut event.instructions,
                 );
+
+                // Technically, we could still try to apply a pending checkpoint after receiving a
+                // new one. However, sync_local assumes it's only called in a state where there's no
+                // pending checkpoint, so we'd have to take the oplog state at the time we've
+                // originally received the validated-but-not-applied checkpoint. This is likely not
+                // something worth doing.
                 self.validated_but_not_applied = None;
                 *target = updated_target;
             }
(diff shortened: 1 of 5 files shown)


crates/core/src/sync/streaming_sync.rs
```diff
@@ -448,6 +448,12 @@ impl StreamingSyncIteration {
                     |s| s.start_tracking_checkpoint(progress, subscription_state),
                     &mut event.instructions,
                 );
+
+                // Technically, we could still try to apply a pending checkpoint after receiving a
+                // new one. However, sync_local assumes it's only called in a state where there's no
+                // pending checkpoint, so we'd have to take the oplog state at the time we've
+                // originally received the validated-but-not-applied checkpoint. This is likely not
+                // something worth doing.
```
rkistner, line 456: > This is likely not something worth doing.

The current implementation effectively "invalidates" any past checkpoint whenever we receive new data - we do not keep the data around to apply a past checkpoint. While we could implement something to keep the last checkpoint around, it is a big project, so I agree with "likely not worth doing" at least for now.

rkistner (approved):

PR #138, opened 2025-10-15
Build independent shared libraries for Android

simolus3:
This is the Android follow-up to https://github.com/powersync-ja/powersync-sqlite-core/pull/135. Because 36 attachments to each release are apparently not enough.

The motivation is the same: For Dart build hooks, we want to be able to download a `.so` for each Android ABI independently instead of depending on a Gradle buildscript.
Because the file names clash with Linux builds, this also renames Linux binaries to `.linux.so` (used to just be `.so`).


rkistner (approved):
Happy with adding these.

I do wonder - could it make sense to zip together related artifacts? For example all Android so files in one zip, or even all dynamic libraries in one zip?

PR #151, opened 2025-11-25
feat: add app_metadata to rust sync requests

stevensJourney:
# Overview

This passes `app_metadata` provided by the client to Sync stream request payloads.


stevensJourney:
> We need someone else to approve this now, but I'm happy with these changes 👍
> 
> Should we wait for the service changes to land first?

Thanks! 

Indeed, this can wait for the service changes, however if we are planning on doing a core release before those service changes are released - I think we could progress with this. I can't really imagine many other methods of passing this metadata to the service.

PR #172, opened 2026-03-30
Restructure static libraries attached to releases

simolus3:
In this repository, "static" is a bit of an overloaded term:

1. It can refer to whether the core extension library itself is built as a static library.
2. It can refer to whether the core extension invokes SQLite APIs dynamically (by being given a bunch of function pointers at runtime) or statically (by invoking functions directly).

This gives us four possibilities, all of which have their use:

1. Static core extension library + static SQLite invocations: This is used by the `libpowersync-wasm.a` build for the Dart SDK and watchOS builds in the XCFramework. Even though it doesn't break anything, static SQLite invocations for watchOS was likely a mistake. It's only used in the Kotlin and Swift SDK, but approach 2 would be better for that. We also build Android and Desktop variants of this build, but those aren't used. 
2. Static core extension library + dynamic SQLite invocations: This is not currently used, but it would be very useful to have that in the Kotlin SDK! We can statically link external code in the Kotlin SDK, and this build would enable us to link the core extension while still supporting arbitrary SQLite versions (in particular, both sqlite3 and sqlite3multipleciphers).
3. Dynamic core extension library + static SQLite invocations. We use this for the WA-sqlite WASM side-module.
4. Dynamic core extension library + dynamic SQLite invocations: We use this to load the core extension in the Dart, React Native, Node, Capacitor, Kotlin (except for watchOS), C# and Swift SDKs.

This PR removes a few unused builds and prepares statically-linked libraries with dynamic SQLite invocations for Kotlin:

1. The `.a` and `.lib` files attached to our releases use build type 1. They were meant to be linked statically into Dart applications once Dart supports that, but it's unclear whether Dart will ever add support for that. Also, they should have probably used build type 2 instead so that we can still let the app developer decide how to link SQLite .
2. This removes static core extension builds for Android, and adds static core extension builds with dynamic SQLite invocations for all targets we support with Kotlin/Native. This allows linking the core extension from Kotlin.

Linking the core extension through Kotlin makes the SDK easier to use (since users don't have to add an additional native dependency), and unblocks native Linux and Windows support for the Kotlin SDK. I've manually tested this setup in the Kotlin SDK.

### DIFF
--- a/.github/actions/android/action.yml
+++ b/.github/actions/android/action.yml
@@ -61,26 +61,18 @@ runs:
         path: android/build/distributions/powersync_android.zip
         if-no-files-found: error
 
-    - name: Copy static libraries
+    - name: Copy individual libraries
       shell: bash
       run: |
-        cp target/aarch64-linux-android/release/libpowersync.a libpowersync_aarch64.android.a
         cp target/aarch64-linux-android/release/libpowersync.so libpowersync_aarch64.android.so
-
-        cp target/armv7-linux-androideabi/release/libpowersync.a libpowersync_armv7.android.a
         cp target/armv7-linux-androideabi/release/libpowersync.so libpowersync_armv7.android.so
-
-        cp target/i686-linux-android/release/libpowersync.a libpowersync_x86.android.a
         cp target/i686-linux-android/release/libpowersync.so libpowersync_x86.android.so
-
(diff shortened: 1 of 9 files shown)


rkistner (commented):
The description makes sense

rkistner (approved):

PR #184, opened 2026-05-29
Update Rust to 1.96.0

simolus3:
Rust 1.96.0 has been released, and stabilizes `assert_matches!`. This updates the toolchain and adopts that macro (we've used it before, but later removed our usages to support stable Rust).

Rust stopped passing `-Wl,--allow-undefined` to the wasm linker by default, which means that we get linker errors about undefined sqlite symbols now. We intentionally want that for the static library build though, as the SDK (or `wa-sqlite` fork) will link that in the end. We could add explicit wasm import annotations to these functions, but manually passing the linker arg seems easier.

### DIFF
--- a/tool/build_wasm.sh
+++ b/tool/build_wasm.sh
@@ -32,13 +32,13 @@ cp "target/wasm32-unknown-emscripten/wasm_asyncify/powersync.wasm" "libpowersync
 # Static lib.
 # Works for both sync and asyncify builds.
 # Works for both emscripten and wasi.
-# target/wasm32-wasip1/wasm/libpowersync.a
+# target/wasm32-unknown-unknown/wasm/libpowersync.a
 cargo build \
   -p powersync_loadable \
   --profile wasm \
   --no-default-features \
   --features "static nightly" \
   -Z build-std=panic_abort,core,alloc \
-  --target wasm32-wasip1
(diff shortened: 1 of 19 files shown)


tool/build_wasm.sh
```diff
@@ -32,13 +32,13 @@ cp "target/wasm32-unknown-emscripten/wasm_asyncify/powersync.wasm" "libpowersync
...
-# target/wasm32-wasip1/wasm/libpowersync.a
+# target/wasm32-unknown-unknown/wasm/libpowersync.a
 cargo build \
   -p powersync_loadable \
   --profile wasm \
   --no-default-features \
   --features "static nightly" \
   -Z build-std=panic_abort,core,alloc \
-  --target wasm32-wasip1
+  --target wasm32-unknown-unknown
```
simolus3, line 42: I don't know why we would target WASI here. We have a `no_std` build so it kind of doesn't matter, but `package:sqlite3` definitely doesn't provide necessary imports. `wasm32-unknown-unknown` is the "no external imports" target and seems to work here.

I've pulled this static library into the PowerSync Dart SDK locally and ran tests with it. Most of them failed because of the datetime accuracy changes in v0.5.0, but at least that confirms the library is still linkable.

rkistner (approved):

PR #189, opened 2026-06-12
Helpers to move between raw tables and ps_untyped

simolus3:
With the `_rest` column pattern and `powersync_trigger_resync`, we have pretty decent tools integrating PowerSync into migrations altering existing tables.

However, steps to create and drop raw tables could still be simplified a bit:

1. To add raw tables, we currently suggest a [manual copy](https://docs.powersync.com/client-sdks/advanced/raw-tables#adding-raw-tables-as-a-new-table) from `ps_untyped` into the raw table. We forgot to mention that moved rows should also be removed from `ps_untyped` afterwards (https://github.com/powersync-ja/powersync-kotlin/issues/355).
2. For symmetry, one should also move rows back into `ps_untyped` before dropping raw tables. We have no documentation on this at all.

Both steps can be automated fairly easily: For 1, we can use the exact same logic we use in a `sync_local` step to parse the oplog JSON structure from `ps_untyped` through the (inferred or explicit) `put` statement, which copies it into the raw table.
For 2, we can use the existing logic we have to create crud triggers to JSON-ify table contents and write them back into `ps_untyped`.

### Can't we do this implicitly?

[Initially](https://discord.com/channels/1138230179878154300/1504213613009834145/1511830197329199121), I thought we could this in `powersync_init()` / `powersync_update_schema()`, just like we auto-migrate views:

1. We could detect a new raw table in the schema and automatically copy data over.
2. ... and that's where it stops being a good idea, because we can't detect deleted raw tables missing from the schema.

Even 1 breaks in practice because depending on the SDK, users might first call `powersync_init` with the raw table already in the schema before using `execute()` APIs in SDKs to actually create the raw table.

So, this PR adds an explicit function for this, `powersync_raw_table_migrate`. It's supposed to be called like this:

- For new raw tables, after the table has been created and added to the app's schema. Invoke it with `powersync_raw_table_migrate('create', 'name_of_raw_table_in_sync_streams')` to move data from `ps_untyped`.
- For dropped raw tables, before the table gets dropped in a migration. The table would not be part of the app's schema at this point, so it takes a JSON description of the table (necessary to describe local-only columns, which should not be moved to `ps_untyped`).

_AI use disclaimer_: Parts of the implementation and most tests were generated with Claude.

### DIFF
--- a/crates/core/src/schema/mod.rs
+++ b/crates/core/src/schema/mod.rs
@@ -4,7 +4,9 @@ mod management;
 mod raw_table;
 mod table_info;
 
-use alloc::{rc::Rc, vec::Vec};
+use core::{ffi::c_void, fmt::Write};
+
+use alloc::{format, rc::Rc, vec::Vec};
 pub use common::{ColumnFilter, SchemaTable};
 use powersync_sqlite_nostd::{self as sqlite, Connection, Context, Value, args};
 pub use raw_table::InferredSchemaCache;
@@ -17,9 +19,12 @@ pub use table_info::{
 
 use crate::{
     error::{PSResult, PowerSyncError},
-    schema::raw_table::generate_raw_table_trigger,
+    ext::SafeManagedStmt,
+    schema::raw_table::{InferredTableStructure, generate_raw_table_trigger},
     state::DatabaseState,
-    utils::WriteType,
+    sync::RawTableWithCachedStatements,
+    utils::{SqlBuffer, WriteType},
+    views::table_columns_to_json_object,
 };
 
 #[derive(Deserialize, Default)]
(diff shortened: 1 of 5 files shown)


rkistner:
I wonder if we should rather remove `ps_untyped` completely at some point, and just rely on `powersync_trigger_resync` syncing the changes on the next checkpoint? It would be a breaking change though.

Generally, there are two scenarios when a new table is added in a new app update:
1. The app syncs the new table on the old app version, then the user updates the app version. `ps_untyped` + the `_rest` column (for raw tables) handles the migration when the new app version is launched. This is nice, since the data is available immediately.
2. The user updates the app version first, then syncs the new table. In this case, there is no way to have the data for the new table available before completing a sync.

So I'm wondering, is it worth all the additional complexity and overhead to support scenario 1, when the app still has to deal with the data only becoming available after a sync in scenario 2? Versus just always handling migrations as scenario 2, requiring syncing a new checkpoint after creating the table?

I added some additional notes on migration strategies in [this doc](https://docs.google.com/document/d/1h2sayKHsQ2hwSAaBlR8z7ReEVDeJ2t1Zzs8N0um5SJc/edit?tab=t.0) a while back.

I haven't thought through all the implications of completely removing `ps_untyped` though, especially for JSON-based tables.

simolus3:
> So I'm wondering, is it worth all the additional complexity and overhead to support scenario 1, when the app still has to deal with the data only becoming available after a sync in scenario 2

Good point, I agree that using a resync after migrations sounds like a more reliable tradeoff. With checkpoint requests, knowing when data is ready after a migration should also be pretty straightforward (or one just lets `powersync_trigger_resync(TRUE)` clear all progress).

simolus3:
> I haven't thought through all the implications of completely removing `ps_untyped` though, especially for JSON-based tables.

I assume that raw table users would in many cases have a schema that only consists of raw tables without any JSON views. So we could potentially have a schema option to disable `ps_untyped`, and recommend enabling it for users relying on raw tables.

We can then still enable it by default later if we want to make that breaking change.

PR #198, opened 2026-06-29
feat: checkpoint requests

stevensJourney:
# Overview

This implements the core changes for the Checkpoint Requests proposals:

- https://github.com/orgs/powersync-ja/discussions/324
- https://github.com/orgs/powersync-ja/discussions/317

The proposals refactor write checkpoints into a general Checkpoint Requests methodology: clients
now generate and track checkpoint request ids themselves (previously the PowerSync service
generated write checkpoint ids). The service treats client-generated checkpoint request ids as the
write checkpoint ids it reports to older-protocol clients, so both flows share one id namespace.

For background on the legacy system, see `docs/historic-write-checkpoints.md`. For a detailed
overview of the new system, see `docs/write-checkpoint-requests.md`. The `powersync_control`
command and instruction reference in `docs/sync.md` has been updated accordingly.

# Migrations

Write checkpoint state was previously tracked in a synthetic `$local` bucket in `ps_buckets`.
Migration **v14** moves this state into dedicated `ps_kv` keys:

| Legacy (`$local`) | New (`ps_kv`)                        |
| ----------------- | ------------------------------------ |
| `last_applied_op` | `last_applied_checkpoint_request_id` |
| `last_op`         | `last_seen_checkpoint_request_id`    |
| `target_op`       | `target_checkpoint_request_id`       |

The migration then deletes the `$local` row (so `ps_buckets` only contains real sync buckets) and
drops the `ps_buckets.target_op` column, so older SDKs fail hard rather than silently diverge if
they open a migrated database. A full down migration restores the v13 schema and rebuilds the
`$local` row from `ps_kv`; re-upgrading after a downgrade treats the `$local` row as the source of
truth, including any progress an older SDK made in between. A read/write audit confirmed the mapped
`ps_kv` usages are semantically identical to the legacy `$local` bookkeeping.

A new `last_requested_checkpoint_request_id` key holds the client-side allocation counter. It is
intentionally not seeded from migrated state: SDKs reconcile the counter with the service on every
connect (see below). See the "Migration from `$local`" section in
`docs/write-checkpoint-requests.md` for details, including the ambiguous max-op-sentinel case.

# Additions

All new functionality is exposed through `powersync_control`, consistent with the rest of the sync
interface. Unlike most `powersync_control` commands, the three query/allocation commands below
return their result directly as a SQLite value instead of an instruction array — they are
responses to a function call, not sync-lifecycle instructions:

- **`next_checkpoint_request_id`** — allocates, persists and returns the next checkpoint request
  id as an integer result. Requires an active sync iteration and seeded request state. It
  participates in the caller's transaction, so a rolled-back allocation can be retried safely.
- **`current_checkpoint_request_id`** — returns the current sequence value (or `NULL` when
  unseeded) without allocating. SDKs use this to repost an existing checkpoint request after a
  crash or restart without burning a new id.
- **`target_checkpoint_request_id`** — probes, updates, or clears (payload `0`) the target used to
  block applying downloaded rows while local writes are outstanding, returning the previous value.
  Works outside a sync iteration. This is the split that keeps explicit checkpoint requests from
  blocking applies: only write checkpoints update the apply gate.
- **`seed_checkpoint_request_id`** — seeds the allocation counter from service state. Requires an
  active sync iteration. Rejects `NULL` and `0` payloads: a conforming SDK always has a positive
  id to seed after reconciliation.

The query/allocation commands accept the request id as either a 64-bit integer or a decimal string
(the string form lets JavaScript SQLite drivers pass values above `2^53` without precision loss).

## Enabling checkpoint requests and seeding

Checkpoint requests are opt-in per iteration: the SDK sets `checkpoint_mode: "requests"` on
`StartSyncStream` (the default is `legacy`). When enabled, `EstablishSyncStream` carries an optional
`checkpoint_request` payload:

```json
{ "client_id": "<client id>", "checkpoint_request_id": "<decimal string>" }
```

`checkpoint_request_id` is core's initial request id — the maximum of the local allocation counter,
any concrete local-write target, and `1` — serialized as a decimal string to preserve the full
64-bit range in JSON clients. The payload is omitted entirely in legacy mode.

On every connection attempt, the SDK reconciles this value with the service's checkpoint-request
state (posting at least `1`) and seeds core with the response via `seed_checkpoint_request_id`. Core
stores seeded values **verbatim** — it does not enforce monotonicity. The SDK owns id validity and
cannot allocate new requests until seeding completes. This connect-time reconciliation doubles as a
re-request, so checkpoint-request waiters resolve in every new session without durable applied state
in core.

## Notifying applied checkpoint requests

When a full checkpoint carrying a `write_checkpoint` is applied locally, core reports the applied
request id through two channels:

- **`DidCompleteSync.applied_checkpoint_request_id`** — the canonical signal. The field is only
  serialized when the applied checkpoint carried a write checkpoint (so the instruction is
  unchanged for other checkpoints), and is encoded as a decimal string for JSON precision.
- **`UpdateSyncStatus.status.internal_last_applied_checkpoint_request_id`** — an event-style field
  (also a decimal string) for SDKs that drive waiters from the sync status stream (e.g. web-worker
  setups where reacting to status is easier than plumbing an instruction through). It is set on the
  status update that follows a full-checkpoint apply, overwritten by the next full apply (or cleared
  when that apply has no request id), and cleared on (re)connect and disconnect. It is not reset by
  intermediate download-progress or partial-completion status updates, so an SDK should treat a
  carried id idempotently (act on it once per id) rather than as a fresh event on every status
  snapshot. It is runtime-only SDK state, not app-visible progress, and is never reported by
  `powersync_offline_sync_status`.

Partial (priority) checkpoint completions never report an applied request id and never advance the
seen/applied state.

A general SDK `requestCheckpoint` flow:

1. Allocate an id: `powersync_control('next_checkpoint_request_id', NULL)` (in a transaction).
2. Post the id to the service's checkpoint request endpoint.
3. For write checkpoints only, set the apply gate:
   `powersync_control('target_checkpoint_request_id', id)`.
4. Wait for a `DidCompleteSync` instruction (or status update) with an applied id `>= id`.

Local writes set `target_checkpoint_request_id` to the max-op sentinel and clear the seen/applied
high-water marks, preserving the legacy behavior that only checkpoint request ids observed *after*
a write can satisfy the apply gate.

# Testing

- Dart tests cover the new control commands (including direct return values and payload
  validation), seeding semantics (verbatim/non-monotonic stores, `NULL`/`0` rejection), apply-gate
  behavior, and applied-id reporting through both `DidCompleteSync` and the sync status for full
  vs. partial checkpoints.
- Migration tests cover up, down, and re-upgrade-after-downgrade paths, including sentinel target
  ops and schema-equality assertions against fixtures.

Related PRs for the PowerSync service and the initial Swift SDK implementation will be linked here.

<details>
<summary>Historical context and alternatives considered</summary>

## Alternatives considered

- **Standalone SQLite functions** — the initial implementation exposed
  `powersync_next_checkpoint_request_id()` and `powersync_probe_local_target_op()` as dedicated
  SQLite functions, e.g.:

  ```swift
  let requestId = try await db.writeTransaction { ctx in
      try ctx.get(sql: "SELECT powersync_next_checkpoint_request_id()", parameters: []) { cursor in
          try cursor.getInt64(index: 0)
      }
  }
  ```

  These were replaced with `powersync_control` commands so the checkpoint request lifecycle lives
  in the same interface (and connection) as the rest of the sync client.

- **Instruction-shaped responses** — after the move to `powersync_control`, the query/allocation
  commands initially returned dedicated `CheckpointRequestId` and `LocalTargetOp` instructions.
  These weren't really instructions — they were responses to a function call — and handling them
  required special cases in SDKs. They now return their values directly as SQLite results, handled
  in `control()` without ticking the sync client (mirroring how `subscriptions` is handled).

- **A dedicated `CheckpointRequestApplied` instruction** — applied checkpoint requests were first
  notified through a separate instruction. Since it was only ever emitted together with
  `DidCompleteSync` for a fully-applied checkpoint, it was merged into
  `DidCompleteSync.applied_checkpoint_request_id`, which also removed any ordering question
  between the two instructions.

- **Sync status as the notification stream** — an earlier revision emitted the applied request id
  *only* on the `SyncStatus` stream. That was first dropped in favor of the dedicated instruction
  (to keep an internal high-water mark from being presented as meaningful sync progress), then
  partially reinstated: the status now carries `internal_last_applied_checkpoint_request_id` as an
  event-style, runtime-only field *alongside* the canonical `DidCompleteSync` signal, because some
  SDK architectures (e.g. web workers) find it much easier to react to the status stream than to
  individual instructions. The `internal_` prefix and the clear-on-reconnect semantics keep it
  from being mistaken for persisted state or user-facing progress.

- **`max()` seeding in core** — seeding originally used `max(local, service)` semantics so the
  counter could never move backwards. This was dropped in favor of storing seeded values verbatim:
  the SDK performs the bidirectional reconciliation with the service and is the only party that can
  judge which value is correct (e.g. after switching users or a service-side reset). The same
  reasoning applies to `last_applied_checkpoint_request_id`, which is always the id from the last
  applied checkpoint as sent by the service.

- **Allowing `NULL` seeds** — an earlier revision accepted `NULL` as a seed to mean "no service
  state". Since a conforming SDK always posts at least `1` during reconciliation and seeds the
  accepted response, `NULL` (and `0`) seeds are now rejected by core.

- **Naming: `local_target_op`** — the apply-gate key and control command were originally named
  `local_target_op`, carrying over the legacy `ps_buckets.target_op` terminology. Since `target_op`
  referred to a bucket concept that no longer exists on the client, both were renamed to
  `target_checkpoint_request_id`. The legacy column name survives only in the v13 schema, the down
  migration that restores it, and the historic docs.

- **Single JSON value vs. separate `ps_kv` keys** — the migrated state could have been stored as
  one JSON document under a single key. Separate keys were chosen to keep querying simple (both for
  core's apply-gate SQL and for debugging).

- **Seeding the request counter from migrated `target_op`** — a concrete legacy `$local.target_op`
  could seed `last_requested_checkpoint_request_id` during migration. This was left out as
  redundant: SDKs reconcile the counter with service state on connect before advancing it.

- **Keeping the `target_op` column** — the column could have been retained for compatibility.
  Dropping it was chosen deliberately so that older SDK versions fail with a hard SQLite error on
  local writes instead of silently maintaining state the new implementation no longer reads. The
  down migration restores the column for supported downgrades.

</details>

---
AI Disclosure: I initially implemented the work by hand without AI. Codex 5.5 assisted with
creating tests and writing docs; Claude Code assisted with a review pass, cleanups, and doc
updates. All AI changes have been manually guided and verified.


crates/core/src/sync/interface.rs
```diff
@@ -324,6 +324,28 @@ pub fn register(db: *mut sqlite::sqlite3, state: Rc<DatabaseState>) -> Result<()
         Some(DatabaseState::destroy_rc),
     )?;
 
+    db.create_function_v2(
```
simolus3: Nit / up for discussion: Most of the other functionality related to updating sync client state is exposed through `powersync_control`, so I wonder if `powersync_control('next_checkpoint_request_id', NULL)` and `powersync_control('local_target_op', ?)` might be more consistent?

crates/core/src/sync/storage_adapter.rs
```diff
@@ -500,18 +511,132 @@ WHERE bucket = ?1",
...
+    pub fn next_checkpoint_request_id(&self) -> Result<i64, PowerSyncError> {
+        let statement = self.db.prepare_v2(
+            "INSERT INTO ps_kv(key, value)
+VALUES(?1, 1)
+ON CONFLICT(key) DO UPDATE SET value = CAST(value AS INTEGER) + 1
+RETURNING value",
+        )?;
+        statement.bind_text(
+            1,
+            LAST_REQUESTED_CHECKPOINT_REQUEST_ID_KEY,
```
stevensJourney, line 590: Storing the checkpoint request id on the client side has one edge case that I
think is worth calling out.

The proposed protocol treats checkpoint request ids as client-generated and
incrementing for a `(user_id, client_id)` pair. In the this SQLite core
implementation, the client-side sequence is stored in `ps_kv` as
`last_requested_checkpoint_request_id`. `powersync_clear(...)` currently deletes
all `ps_kv` entries except `client_id`, so a clear also removes this sequence
state.

One scenario:

- There is a large replication lag in the sync service.
- During that lag, the client completes several uploads. After each upload batch,
  it allocates a new checkpoint request id and, once the service accepts the
  request, stores that id as the local `target_op`.
- Assume the latest checkpoint request id accepted by the service for this
  client is `7`.
- The client then runs `disconnectAndClear` after uploads have finished. Since
  clear removes `last_requested_checkpoint_request_id`, the local sequence state
  is lost.
- If the client later performs more mutations and requests another checkpoint,
  the next locally allocated checkpoint request id would be `1`.

From the local sync flow's point of view, I do not immediately see a correctness
problem if the service accepts checkpoint request id `1` and the client stores
`target_op = 1`: the client would wait for a checkpoint associated with the new
replication head. It would, however, mean the service-side checkpoint request id
for the same `(user_id, client_id)` has gone backwards, which may be surprising
in logs and weakens the "incrementing client-generated id" property.

One possible mitigation is the "strictly incrementing checkpoint ids" server-side
alternative from https://github.com/orgs/powersync-ja/discussions/317. That
would address the race condition described in the "Risks / Drawbacks" section,
where an older delayed request can arrive after a newer one and replace it. If
the service receives a checkpoint request id lower than the current value, it
could reject the request and return the current value. A client that lost its
local `last_requested_checkpoint_request_id` could then reseed its local counter
with that value and retry with the next id.

This has the same recovery tradeoff already called out in that alternative: it
only works while the service still has checkpoint request state for that
`(user_id, client_id)`. If the temporary checkpoint request record has expired or
been deleted, the service no longer has a value to return, and the client would
have to start again from `1`.

If we want to avoid losing the client-side sequence on clear, another option is
to move `last_requested_checkpoint_request_id` into a dedicated table that is not
cleared by `powersync_clear(...)`. Since the service scopes checkpoint requests
by `(user_id, client_id)`, that table would likely need to include `user_id`
alongside the last requested checkpoint request id. We could potentially get the
`user_id` (or some hashed version) from the response of the `/sync/checkpoint-request` route.

crates/core/src/sync/interface.rs
```diff
@@ -340,8 +362,74 @@ fn powersync_offline_sync_status_impl(
     Ok(serialized)
 }
 
+fn powersync_probe_local_target_op_impl(
+    ctx: *mut sqlite::context,
+    args: &[*mut sqlite::value],
+) -> Result<Option<i64>, PowerSyncError> {
+    if args.len() != 1 {
```
simolus3: SQLite validates the amount of arguments passed to functions, so we don't need to check for this and can use `args[0]` directly (same in `powersync_next_checkpoint_request_id_impl` below).

crates/core/src/sync/interface.rs
stevensJourney: I think this is a fair point. I was actually wondering about potentially integrating with the sync client's core state today. My at-the-time thought was that having the internal state machine track the re-seeding of `checkpoint_request_id`s on `connect` could be cleaner. I will definitely investigate this path.

crates/core/src/sync/storage_adapter.rs
stevensJourney: Resolving this: the connect-time reconciliation contract now covers this scenario, and core enforces the recovery path.

Walking the exact scenario with the current implementation:

1. `disconnectAndClear` deletes `last_requested_checkpoint_request_id` (along with the other checkpoint keys). Crucially, that also removes the *seeded* marker — `powersync_control('next_checkpoint_request_id')` fails until the counter has been re-seeded, and seeding only happens through the connect flow. So the client **cannot** allocate id 1 while the service is still at 7; there is no window for the sequence to regress.
2. On the next connect, `EstablishSyncStream` reports `last_checkpoint_request_id: null`. The SDK re-affirms state with the service by posting a checkpoint request with an id of at least 1 (max of the local hint and any concrete local target). The endpoint returns the **effective request state accepted by the backend** — 7 in this scenario, not the 1 the client posted.
3. The SDK seeds core with that response, so the next locally allocated id is 8. The service-side id never goes backwards while the service still has a record.

Step 2 is essentially the "strictly incrementing ids" mitigation from the discussion: the service doesn't let a lower id replace its current state — it answers with the effective value, and the client reseeds from the response. This also resolves the delayed-request race called out in the Risks section.

The remaining tradeoff is the one already accepted in the proposal: if the service's record has *also* expired, both sides have genuinely lost the value and the counter restarts from 1. That's documented as acceptable — the local sync flow stays correct (the apply gate only compares ids observed after the last local write, and seeds are stored verbatim), and "surprising in logs" is the full extent of the impact.

Given that, I don't think the dedicated non-cleared table is needed: the connect flow restores the counter whenever the service remembers it, and losing it on clear is actually desirable for the user-switch case (the reconciliation always runs against the authenticated `(user_id, client_id)`, so scoping is handled implicitly rather than by persisting a `user_id` locally).

This is now written down in `docs/write-checkpoint-requests.md` (the reconciliation section, the `powersync_clear` note under "`ps_kv` checkpoint state", and the "post ≥ 1, seed the response" guidance) and in the `seed_checkpoint_request_id` entry in `docs/sync.md`.

crates/core/src/sync/interface.rs
stevensJourney: 👍 This function no longer exists :)

crates/core/src/sync/interface.rs
stevensJourney: I've updated these to use `powersync_control` invocations now. This did feel like a cleaner solution.

crates/core/src/sync/storage_adapter.rs
```diff
@@ -26,6 +26,15 @@ use super::{
...
 };
 
+const LAST_REQUESTED_CHECKPOINT_REQUEST_ID_KEY: &str = "last_requested_checkpoint_request_id";
+const LAST_SEEN_CHECKPOINT_REQUEST_ID_KEY: &str = "last_seen_checkpoint_request_id";
+const LAST_APPLIED_CHECKPOINT_REQUEST_ID_KEY: &str = "last_applied_checkpoint_request_id";
+
+// Tracks the local target used to block applying downloaded rows while local writes are
+// outstanding. When present, this is normally either the max-op sentinel for pending local writes or
+// a concrete checkpoint request id also stored in LAST_REQUESTED_CHECKPOINT_REQUEST_ID_KEY.
+const LOCAL_TARGET_OP_KEY: &str = "local_target_op";
```
rkistner: Would something like `target_checkpoint_request_id` be a better description here? `target_op` originally referred to a concept for buckets that is not really relevant on the client anymore.

rkistner (commented):
I'm happy with the model overall, but I'll leave review of the technical implementation to @simolus3.

Reading the docs, one part that's not 100% clear to me is `seed_checkpoint_request_id`:
1. When exactly is this meant to be called? _seed_ typically implies it's during setup only, or perhaps once per connection. Or is that potentially after every request to the server? Describing the SDK implementation like for other operations would help a lot here.
2. Are there concurrency risks here? For example: (1) client sends 3x requests to the server, each incremental `last_checkpoint_request_id` (2) client receives a response back from the server, calls `seed_checkpoint_request_id`, which then sets it back to an older value. Current docs mention the client SDK is responsible for reconsiling these, which can include dealing with concurrency issues like that, but the docs should be clear on what the client SDK should do.

On the docs: It's nice to have all the details, and the SDK implementation descriptions are really useful. But some of the other docs are quite verbose and still difficult to follow: It describes a lot of detailed mechanics, but I'm missing the high-level purpose. For example, it took some digging to see what exactly each of the `ps_kv` fields are used for, and why we need the 4x separate fields.

stevensJourney:
> 1. When exactly is this meant to be called? _seed_ typically implies it's during setup only, or perhaps once per connection. Or is that potentially after every request to the server? Describing the SDK implementation like for other operations would help a lot here.

This currently happens once per connection. This is used for the case where the PowerSync service might have a checkpoint record, but the client doesn't - perhaps after a `disconnectAndClear`.

> Are there concurrency risks here? For example: (1) client sends 3x requests to the server, each incremental last_checkpoint_request_id (2) client receives a response back from the server, calls seed_checkpoint_request_id, which then sets it back to an older value. Current docs mention the client SDK is responsible for reconsiling these, which can include dealing with concurrency issues like that, but the docs should be clear on what the client SDK should do.

We limit the concurrency on the SDK level currently. Essentially, we do 1 request to the service on each connect iteration. We wait for the response from the server before allowing other checkpoint-requests from executing. The client's sequence is then seeded from the server response. Subsequent checkpoint requests then use the client's sequencing. 

On the service side, the PowerSync service avoids concurrency issues by only advancing the `checkpoint_request_id` if it's increasing. Clients then receive the larger (potentially service side state) response - which they wait for. 

> On the docs: It's nice to have all the details, and the SDK implementation descriptions are really useful. But some of the other docs are quite verbose and still difficult to follow: It describes a lot of detailed mechanics, but I'm missing the high-level purpose. For example, it took some digging to see what exactly each of the ps_kv fields are used for, and why we need the 4x separate fields.

That's fair feedback. I'll use this to improve the docs.

crates/core/src/sync/interface.rs
```diff
@@ -128,13 +139,29 @@ pub enum Instruction {
...
     FetchCredentials {
         /// Whether the credentials currently used have expired.
         ///
         /// If false, this is a pre-fetch.
         did_expire: bool,
     },
+    /// Return a newly allocated checkpoint request id to the SDK.
+    CheckpointRequestId { request_id: i64 },
+    /// Notify the SDK that a checkpoint request id has been applied locally.
+    CheckpointRequestApplied { request_id: i64 },
```
simolus3: I wonder if we could inline this into `DidCompleteSync`?

crates/core/src/sync/interface.rs
```diff
@@ -128,13 +139,29 @@ pub enum Instruction {
...
+    /// Return the local target op value observed before an optional update.
+    LocalTargetOp { target_op: Option<i64> },
```
simolus3: `CheckpointRequestId` and `LocalTargetOp` aren't really instructions, they're more like a response to a function call, right? Adding them to this enum kind of blurs the types (and this is also visible in the Swift SDK implementation where these two are handled specially). Apart from `EstablishSyncStream`, which arguably has the same problem, instructions are meant as requests that can appear at any point in the sync process.

This could use some restructuring still, but overall I think the control flow would be cleaner if we could invoke this directly in `control()` like e.g. this:

```Rust
"next_checkpoint_request_id" => {
    let client = state.sync_client.borrow();
    let has_sync_iteration = client
        .as_ref()
        .map(|e| e.has_sync_iteration())
        .unwrap_or(false);

    if !has_sync_iteration {
        return Err(PowerSyncError::state_error("No iteration is active"));
    }
    let adapter = state.storage_adapter(db)?;
    if !adapter.has_checkpoint_request_id()? {
        return Err(PowerSyncError::state_error(
            "Checkpoint request state has not been seeded",
        ));
    }

    let request_id = adapter.next_checkpoint_request_id()?;
    ctx.result_int64(request_id);
    return Ok(());
}
```

I don't think there's a need to tick the sync client here, the `subscriptions` command is handled in a similar way for example.

(I see how my earlier suggestion to move these into `powersync_control` calls probably prompted this, sorry for not being clearer there)

crates/core/src/sync/interface.rs
```diff
@@ -345,3 +389,27 @@ create_sqlite_text_fn!(
...
+
+fn parse_optional_i64_payload(
+    payload: *mut sqlite::value,
+    name: &'static str,
+    type_error: &'static str,
+) -> Result<Option<i64>, PowerSyncError> {
+    let value = match payload.value_type() {
+        ColumnType::Null => return Ok(None),
+        ColumnType::Integer => payload.int64(),
+        ColumnType::Text => payload
```
simolus3, line 474: What's the motivation for allowing text values, JavaScript compatibility? I'd hope we could handle that with big ints, but if we want this for safety then a small comment might help :)

crates/core/src/view_admin.rs
```diff
@@ -164,7 +163,9 @@ fn trigger_resync(db: *mut sqlite::sqlite3, state: &DatabaseState) -> Result<(),
         }
     }
 
-    db.exec_safe("UPDATE ps_buckets SET last_applied_op = 0 WHERE name != '$local'")
+    // Since migration v14, ps_buckets only contains real sync buckets: the synthetic `$local`
+    // bucket has moved to ps_kv, so no filter is needed here.
```
simolus3: Nit: We don't need comments to describe what code no longer does ;) I also see a few more references to `$local`, e.g. in `collect_bucket_requests` and comments in `crud_vtab.rs` and `raw_table.rs`. These should probably be removed too?

docs/write-checkpoint-requests.md
```diff
@@ -0,0 +1,381 @@
...
+the same client id. SDKs may also refresh service state when their user/client context changes.
+
+Because seeds are stored verbatim, seeding `NULL` (or a low id) while core holds a higher counter
+resets that counter, and previously allocated ids would be handed out again. SDKs must therefore
+never forward a raw service response without reconciling: the recommended pattern is to always post
+an id of at least `1` (the maximum of the local hint and any concrete local target) during
+reconciliation and seed core with the service's response to that request. In practice SDKs never
+need to seed `NULL` at all — when there is no local record, posting a checkpoint request with id
+`1` works and doubles as a probe of the service's checkpoint-request support. Core still accepts a
+`NULL` seed for completeness.
```
simolus3: I don't really understand this part yet, but IMO if we won't have SDKs seeding with `NULL` then it shouldn't be supported by the core extension.

crates/core/src/sync/interface.rs
simolus3: Also, I wonder if it makes sense to move that to the sync status as a hidden field only used internally in SDKs.

For our SDKs using web workers where it's not that trivial to call methods on the sync client directly, I'm kind of hoping the implementation of `waitForSync` after acquiring a request id in SDKs could be something like:

```
await for (final snapshot of syncStatus) {
  if (snapshot.internalAppliedRequestId >= requestId) return;
  if (snapshot.error case final error?) throw error;
}
```

(but there are probably a lot more cases I'm missing here)

simolus3 (commented):
I agree that more information on checkpoint request seeding and reconciliation would be helpful, the docs say that this is owned by SDKs but having an overview of the process and more details on the "recommended pattern" part would help.

crates/core/src/sync/storage_adapter.rs
stevensJourney: Yeah that does sound nicer. I've renamed the key.

crates/core/src/sync/interface.rs
stevensJourney: Fair point! I don't know how I missed this event already existed and was triggered at the exact same time :face-palm:. I've removed `CheckpointRequestApplied` and added the last applied checkpoint request to  `DidCompleteSync`

crates/core/src/sync/interface.rs
stevensJourney: I see what you mean. Returning the result directly is nicer. I've implemented this now 😄

crates/core/src/view_admin.rs
stevensJourney: Indeed. I've updated these comments.

docs/write-checkpoint-requests.md
stevensJourney: Fair. After going through all the possible use-cases, I confirmed we don't need to clear this value by passing in a NULL. I've removed this functionality.

crates/core/src/sync/interface.rs
stevensJourney: I actually initially tried using the sync status - it's mentioned in the PR description's `Historical context` section. Using the sync status would be a convenient method for passing this information in SDKs - due to the existing infrastructure.  

The at-the-time reason for moving away from the sync status was mainly related to not wanting to expose the `last_applied_checkpoint_request_id` publicly. It felt like exposing it publicly could result in more confusion than gain. Where the main sources of confusion are that:
- The last applied checkpoint request ID cannot really convey progress. Due to the manner we store checkpoint requests on the server, the value will only ever jump to the latest value. 
- The value should not be persisted between connect iterations. There are edge cases where this value could be invalidated between iterations. The `PowersyncDatabase.syncStatus` usually persists state between connect iterations. 

My reasoning was that, although not trivial, it's not extremely difficult to wait and signal checkpoint application for web workers. We need some signalling in the SDK regardless, to wait for checkpoint seeding etc. So, I moved this out from the SyncStatus.

That being said. After having evaluated all the edge cases, I think if we make the `last_applied_checkpoint_request_id` an _internal_ field on the SyncStatus, then we could take advantage of it. Using the SyncStatus does make the implementation simpler. I've added this now to the implementation and the Swift SDK (will update that PR soon).

crates/core/src/sync/interface.rs
stevensJourney: Yeah. The text value was added with that in mind. IIRC I think Capacitor didn't support `Bigint`s. I added a comment.

crates/core/src/sync/storage_adapter.rs
```diff
@@ -26,6 +26,17 @@ use super::{
     bucket_priority::BucketPriority, interface::BucketRequest, streaming_sync::OwnedCheckpoint,
 };
 
+pub(crate) const LAST_REQUESTED_CHECKPOINT_REQUEST_ID_KEY: &str =
+    "last_requested_checkpoint_request_id";
+pub(crate) const LAST_SEEN_CHECKPOINT_REQUEST_ID_KEY: &str = "last_seen_checkpoint_request_id";
+pub(crate) const LAST_APPLIED_CHECKPOINT_REQUEST_ID_KEY: &str =
+    "last_applied_checkpoint_request_id";
+
```
simolus3: Nit: Since these symbols are not exported from `lib.rs` (either through a `pub use` or if they were part of a `pub mod`), the `(crate)` doesn't do anything.

We don't really have a consistent code style for the core extension so I see how this is bikeshedding, but I think not applying it is simpler.

crates/core/src/sync/storage_adapter.rs
```diff
@@ -499,23 +511,146 @@ WHERE bucket = ?1",
...
+    /// Returns the latest checkpoint request id known locally.
+    pub fn last_checkpoint_request_id(&self) -> Result<Option<i64>, PowerSyncError> {
+        self.read_i64_kv(LAST_REQUESTED_CHECKPOINT_REQUEST_ID_KEY)
+    }
+
+    /// Seeds the local checkpoint request counter from service state.
+    ///
+    /// The value is stored verbatim: core does not enforce monotonicity here. SDKs are
+    /// responsible for reconciling their local hint with the service before seeding, and cannot
+    /// allocate new checkpoint request ids until that seeding has completed.
```
simolus3: For my understanding, is it correct that the flow of seeding checkpoint requests is always going to be this:

1. We emit an `EstablishSyncStream` instruction containing the last request id and kicking off a sync iteration.
2. The SDK then calls `target_checkpoint_request_id`  to check for an existing checkpoint request (compare `>0, != maxOpId`).
3. It then computes the maximum of the last request id and the concrete target, and send that to the service to reconcile.
4. If the response differs from the last request id, SDKs finally call this.

If so, it feels like more of this logic should ultimately be in the core extension instead of the SDK.

1. In the `start` call, SDKs could indicate whether they want to use the new requests checkpoint mode.
2. It then seems like we could add an `Option<{requestId: i64, clientId: String}>` field to `EstablishSyncStream`, and run all of the logic needed to craft the initial `CheckpointRequestPayload` for seeding in the core extension?
3. Finally, since the seeding only happens in the context of a sync iteration, couldn't the core extension remember the `lastCheckpointRequestId` it gave out in the initial request and then just silently do nothing here if they're the same? (we might not even have to do that check since updating a kv entry should be cheap enough to do it unconditionally, and `seed_checkpoint_request_id` has no further side-effects).

docs/write-checkpoint-requests.md
```diff
@@ -0,0 +1,175 @@
+# Write Checkpoint State in `ps_kv`
+
+This document describes the checkpoint state used to keep downloaded rows behind local writes until
+the service has observed those writes, while also supporting explicit "wait until synced"
+checkpoint requests.
+
+The state is internal core/SDK bookkeeping. Apps should not read it as user-facing sync progress.
+
+## State Model
```
simolus3, line 9: This is very helpful, thanks for adding this as documentation 👍

crates/core/src/sync/storage_adapter.rs
stevensJourney: That's fair feedback. I've updated this :)

crates/core/src/sync/storage_adapter.rs
stevensJourney: Yes, that is correct. 

I really like the ideas mentioned here. I've implemented 1 and 2 now. 

I might be potentially misunderstanding `3`. We do need to re-check/seed the last checkpoint request id on every new iteration. I will admit this is being very overcautious - it mainly caters for the fact that a BackendConnector can technically return a token for a different user at any point. So if the user changed between iterations, we should "resync" those counters... If this is referring to caching the `lastCheckpointRequestId` in memory instead of reading from `ps_kv`: that could be an optimisation - I tried looking at this briefly, the number of changes it required felt like more effort than it was worth. As mentioned, reading from `ps_kv` is quite cheap and convenient.

crates/core/src/sync/interface.rs
```diff
@@ -345,3 +443,65 @@ create_sqlite_text_fn!(
...
+    }
+
+    Ok(Some(value))
+}
+
+fn parse_positive_i64_payload(
+    payload: *mut sqlite::value,
+    name: &'static str,
+    type_error: &'static str,
+) -> Result<i64, PowerSyncError> {
```
simolus3: Optional but might be neat: Rust has a dedicated [`NonZeroI64` type](https://doc.rust-lang.org/core/num/type.NonZeroI64.html). That mainly enables memory optimizations (an `Option<NonZeroI64>` is encoded as an `i64` internally), but it's also convenient to express on a type-level that we know a number can't be zero.

So something like `NonZeroI64::new(value).ok_or_else(|| PowerSyncError::argument_error(format!("{name} must be a positive integer")))` might be a more idiomatic way of expressing this, but it also doesn't unlock anything here.

dart/test/sync_test.dart
```diff
@@ -201,6 +273,58 @@ void _syncTests<T>({
...
+      'checkpoint_request_id': '7',
+    });
+  });
+
+  syncTest('omits initial checkpoint request state in legacy mode', (_) {
+    final instructions = invokeControlRaw('start', null);
+    final establish = instructions.whereType<Map>().firstWhere((instruction) =>
+        instruction
+            .containsKey('EstablishSyncStream'))['EstablishSyncStream'] as Map;
+    expect(establish, isNot(containsPair('checkpoint_request', anything)));
```
simolus3: nit: `contains` should match keys on maps


```suggestion
    expect(establish, isNot(contains('checkpoint_request')));
```

simolus3 (dismissed):

crates/core/src/sync/interface.rs
stevensJourney: That does look neater. Thanks for the tip. I've used `NonZeroI64` now.

simolus3 (dismissed):

simolus3 (approved):

PR #199, opened 2026-06-29
Remove RNQS from release checklist template

simolus3:
Since op-sqlite as part of `@powersync/react-native` will be the only option going forward, this removes mentions of the RNQS fork and a separate op-sqlite package from the release checklist.


Chriztiaan (approved):

Chriztiaan:
Do we mark https://github.com/powersync-ja/react-native-quick-sqlite as deprecated?

simolus3:
We should deprecate the package and archive the repository, yes. I think we should do that after the v2 release though, the internal proposal document contains a checklist.

PR #205, opened 2026-07-27
chore: fix release

stevensJourney:
The iOS publish job is currently [failing](https://github.com/powersync-ja/powersync-sqlite-core/actions/runs/30261574160/job/89967101038) with: 

```
release not found
```

When attempting to upload the XCFramework to the draft Github release.

It seems like the iOS workflow does not have the `write` permission configured. Which is needed to upload files to a release.

AI usage: The fix was found by Codex GPT 5.6

The `v0.5.0` draft release currently does not include iOS frameworks. After merging this, we could delete the release and tag and re-push a new tag.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -125,6 +125,8 @@ jobs:
     name: Publish iOS
     needs: [ draft_release, libs_xcframework ]
     runs-on: macos-latest
+    permissions:
+      contents: write # Needed to upload the XCFramework to the draft release.
     steps:
       - uses: actions/checkout@v7
         with:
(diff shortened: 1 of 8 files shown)


simolus3 (approved):
Retagging sounds good to me since the release is not public yet 👍

simolus3 (changes requested):
I just remembered that the release also creates a release to Maven Central, which has already happened for 0.5.0.

So we can't release 0.5.0 again; I see two options here:

1. Manually attach an XCFramework from the CI run (it's attached via `upload-artifact`) to the release as a one-off fix.
2. Bump the version number to 0.5.1 and release again.

We could potentially also do both, with would give us a way to release https://github.com/powersync-ja/powersync-sqlite-core/pull/204 in a 0.5.1 too.

stevensJourney:
> * Manually attach an XCFramework from the CI run (it's attached via `upload-artifact`) to the release as a one-off fix.
> * Bump the version number to 0.5.1 and release again.

Yeah, let's do this.

simolus3 (approved):

simolus3 (approved):

PR #207, opened 2026-07-29
Install pre-close vtab on readonly connections

simolus3:
Some functions using `StorageAdapter` state may reasonably be called on read-only connections:

1. `powersync_offline_sync_status()`  is read-only, and uses the adapter since https://github.com/powersync-ja/powersync-sqlite-core/pull/183.
2. `powersync_control('target_checkpoint_request_id', null)` (added in 0.5.1) is read-only, and also uses the adapter.

Using the storage adapter is generally a good thing because it caches prepared statements. However, these statements need to be finalized before the database can fully close. For write connections, SDKs call `powersync_init()` which registers a custom vtab for this purpose. This function is not called on read-only connections though, so these statements leak and the connection is not properly closed.

As a simple fix that won't require SDK changes, this installs the pre-close hook only when it's needed to clean something up: When hooks are installed or when the storage adapter is created. For this to work on read-only connections, the vtab is now based on `SELECT` statements as `INSERT`s, even into virtual tables, are forbidden.

AI use: Migrating the vtab was done by Claude Code.

### DIFF
--- a/crates/core/src/pre_close_vtab.rs
+++ b/crates/core/src/pre_close_vtab.rs
@@ -140,6 +140,11 @@ static MODULE: sqlite::module = sqlite::module {
     xIntegrity: None,
 };
 
+pub fn ensure_has_internal_close_vtab(db: *mut sqlite::sqlite3) -> Result<(), ResultCode> {
+    db.exec(c"SELECT 1 FROM powersync_internal_close;")?;
+    Ok(())
+}
+
 pub fn register(db: *mut sqlite::sqlite3, state: Rc<DatabaseState>) -> Result<(), ResultCode> {
     db.create_module_v2(
         "powersync_internal_close",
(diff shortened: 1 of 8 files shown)


rkistner (approved):
The approach sounds good to me. Just want to check: How often would `ensure_has_internal_close_vtab` be called now? Just to confirm we're not effectively just preparing `SELECT 1 FROM powersync_internal_close` on every call now, which would kinda defeat the purpose of using prepared statements. It _looks_ like it's called at most once per connection, but just want to confirm.

simolus3:
> Just to confirm we're not effectively just preparing `SELECT 1 FROM powersync_internal_close` on every call now, which would kinda defeat the purpose of using prepared statements

It's just done once per connection, I also confirmed this with breakpoints to be sure. The only call site that creates a new storage adapter is `DatabaseState::storage_adapter`, and the only way for for the adapter to be dropped is for `release_resources` to be called, which only happens when the connection is closed.

PR #211, opened 2026-08-05
Improve error handling

simolus3:
In https://github.com/powersync-ja/powersync.dart/issues/453#issuecomment-5170612953, we've gotten a report where the core extension reports an error like this (emphasis is the part from the extension):

> SqliteException(1): while selecting from statement, **powersync_control: internal SQLite call returned ERROR**, SQL logic error (code 1)
Causing statement: SELECT CAST(powersync_control(?, ?) AS TEXT), parameters: line_binary, blob (55 bytes)

That doesn't help much to understand the problem: Some statement ran by `powersync_control` has failed, but which one?

This PR removes the `From<ResultCode>` implementation for `PowerSyncError`, forcing us to provide more specific details when returning errors. Essentially, this statically prevents forgetting to call `into_db_result`. But even `into_db_result`, which adds more context into errors, wouldn't include the failed statement. So, this adds our own database (`Database`, replacing `*mut sqlite3`) and statement (`Statement`, replacing `ManagedStmt`) wrappers which, instead of returning a `ResultCode` error, will automatically return a `PowerSyncError` with appropriate context and the causing SQL statement.

These wrappers also simplify some things:

1. Instead of returning a `ResultCode` in the `Ok` case, `step()` now returns a `Result<bool>` indicating whether a row is available.
2. Previously, calling `db.exec` with a `&str` would allocate a `CString` to pass to SQLite. This replaces the method with something taking a `&CStr` directly (avoiding a copy for literals).
3. Allow all `PowerSyncError`s to be wrapped with additional context, this was previously only possible for `ResultCode`-backed errors.

AI use: The new helpers are manually written, migrating the rest of the crate was done mechanically with string replacements and AI.
I also ran this through a Code Review with Claude Code to confirm existing behavior is unchanged.


stevensJourney (approved):
This looks like a nice improvement. Changes look good to me.

PR #212, opened 2026-08-11
Add SBOM generator, attest artifacts

simolus3:
This generates an SBOM describing which Rust crates and versons contributed to compiled artifacts, and configures [attestations](https://github.com/powersync-ja/powersync-sqlite-core/attestations) allowing users to verify the provenance of a downloaded core extension file. 

## Build refactoring

To be able to attest builds, they need to run in a reusable workflow [^1]. Currently, we build binaries in composite actions we then reference from both `build.yml` and `release.yml`. This changes the build to happen in a dedicated `build.yml` workflow, triggered through workflow calls from the main workflows. This copies the prior steps, the only exception is the Android build where GPG keys are now passed through with proper secrets instead of as regular inputs.

For releases, the build workflow generates an attestation that binaries were generated by our workflow. Users will be able to run `gh attestation verify -o powersync-ja <their downloaded core extension>` to verify this. We also create an attestation containing an SBOM, which can be used to reconstruct Rust dependencies from the final library, framework or WebAssembly file.

## SBOM generator

I have tried several SBOM generators, but none were a great fit for the core extension:

1. [cdxgen](https://github.com/cdxgen/cdxgen) supports a lot of sources including Cargo, but doesn't work great in workspaces and only looks at `Cargo.toml` or `Cargo.lock`. Only looking at the former misses transitive dependencies, scanning the latter without pruning includes development dependencies that aren't part of the final binary.
2. [cargo-sbom](https://github.com/psastras/sbom-rs) supports workspaces, but also doesn't have a way to remove dependencies that aren't linked into the final binary.
3. [cargo-cyclonedx](https://github.com/cyclonedx/cyclonedx-rust-cargo) is great, but it includes `proc-macro` dependencies and generates an old format (1.5 is the latest version it supports, the latest version of the spec is 1.7). 

Given that the spec for CycloneDX SBOMs is simple and [well documented](https://cyclonedx.org/docs/1.7/json/), this adds a small SBOM generator written in Dart:

- [`cargo tree`](https://doc.rust-lang.org/cargo/commands/cargo-tree.html) is used to crawl actual runtime dependencies, skipping build and proc-macro crates.
- Using names, versions, repositories and license information that we can get directly out of `cargo tree`, this parses a dependency graph and uses that to generate the SBOM.
- Additionally, it adds SQLite as an external component that isn't linked but required at runtime (via `isExternal: true`).

The generated SBOM is attached to our releases.

AI use: CI refactoring is mostly copy-and-paste, using Claude to update the main workflows. I have also used Claude Code to write the SBOM generator (with careful guidance, I manually verified that cyclonedx.studio and grype can parse generated SBOMs).

[^1]: https://docs.github.com/en/actions/how-tos/secure-your-work/use-artifact-attestations/increase-security-rating#building-with-a-reusable-workflow


rkistner (commented):
I didn't check the workflow changes in detail, but the approach and SBOM output look good.

rkistner (approved):

